PR #24, opened 2022-08-31
Unit tests for nomic client and edge case fixes

AndriyMulyar:
Running `make test` after being logged in will run a suite of unit tests. These tests are not yet automated on circleci but do cover several edge cases we have hit in prod users putting in data.

### DIFF
--- a/nomic/atlas.py
+++ b/nomic/atlas.py
@@ -326,13 +326,15 @@ class AtlasClient:
             msg = 'Cannot add embedding to project with modality: {}'.format(project['modality'])
             raise ValueError(msg)
 
+        progressive = len(project['atlas_indices']) > 0
         try:
             self._validate_user_supplied_metadata(data=data, project=project)
         except BaseException as e:
-            self.delete_project(project_id=project_id)
+            if not progressive:
+                self.delete_project(project_id=project_id)
             raise e
 
-        progressive = len(project['atlas_indices']) > 0
+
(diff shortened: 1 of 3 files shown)


nomic/atlas.py
```diff
@@ -239,6 +240,61 @@ def _get_index_job(self, job_id: str):
...
+
+                if project['modality'] == 'text':
+                    if isinstance(datum[key], str) and len(datum[key]) == 0:
+                        msg = 'Datum {} had an empty string for key: {}'.format(datum, key)
+                        raise ValueError(msg)
+
+                if not isinstance(datum[key], (str, float, int)):
+                    raise Exception(f"Metadata sent to Atlas must be a flat dictionary. Values must be strings, floats or ints. Key `{key}` of datum {str(datum)} is in violation.")
+
+        if added_id_field_for_user:
```
bstadt, line 292: this might get loud

nomic/atlas.py
```diff
@@ -271,6 +327,14 @@ def add_embeddings(
             raise ValueError(msg)
 
         progressive = len(project['atlas_indices']) > 0
+        try:
+            self._validate_user_supplied_metadata(data=data, project=project)
+        except BaseException as e:
+            if not progressive:
+                self.delete_project(project_id=project_id)
```
bstadt, line 334: not sure we want to do this here. may be confusing if a user is not going through the map_embeddings interface

bstadt (approved):

PR #31, opened 2022-09-13
Allow empty fields for text projects

AndriyMulyar:


nomic/atlas.py
```diff
@@ -526,7 +530,7 @@ def create_index(self, project_id: str, index_name: str, indexed_field=None, col
         to_return['project_id'] = project['id']
         return CreateIndexResponse(**to_return)
 
-    def add_text(self, project_id: str, data: List[Dict], shard_size=1000, num_workers=10, pbar=None):
+    def add_text(self, project_id: str, data: List[Dict], shard_size=1000, num_workers=10, replace_empty_values_with_null=True, pbar=None):
```
bstadt, line 533: can we make this replace_empty_strings_with_string_null? as its written its confusing.

nomic/atlas.py
```diff
@@ -536,6 +540,7 @@ def add_text(self, project_id: str, data: List[Dict], shard_size=1000, num_worke
         * **data** - An [N,] element list of dictionaries containing metadata for each embedding.
         * **shard_size** - Embeddings are uploaded in parallel by many threads. Adjust the number of embeddings to upload by each worker.
         * **num_workers** - The number of worker threads to upload embeddings with.
+        * **replace_empty_values_with_null** - Replaces empty values in metadata with null. If false, will fail if empty values are supplied.
```
bstadt, line 543: documentations should specify that empty value is a zero length string, and the replacement will be a string with content null

bstadt (approved):

PR #49, opened 2022-10-30
Clarify data shape & provide pandas method suggestion; fix a typo

davidnmora:
When I first tried installing Atlas & running it with my data, & had to hunt around to figure out how to shape my data.

Given `pandas` is basically the default for python data science, providing clarification on the pandas method could help, too.

These small frictions make a big difference when first getting started!


AndriyMulyar:
Thanks for the feedback. I do think this is a workflow where specific documentation for it would be useful. Will review, add more documentation into your PR and then merge.

AndriyMulyar (approved):

PR #62, opened 2022-12-16
Specify topic label estimation field in map_embeddings.

AndriyMulyar:


bmschmidt (approved):

RLesser:
Just a heads up - the documentation states that `topic_label_field` is exposed in AtlasClient, but the latest version (v1.0.28) doesn't include it as an option

PR #72, opened 2023-02-08
Client Refactor

AndriyMulyar:
AtlasProject controls interaction with a project.

The documentation is missing *extensive* documentation on all the ways to use AtlasProject. These will come.


nomic/atlas.py
```diff
@@ -2,1315 +2,234 @@
...
-        None, description="A link to the map this index creates. May take some time to be ready so check the job state."
+def map_embeddings(
+    embeddings: np.array,
+    data: List[Dict] = None,
+    id_field: str = None,
+    map_name: str = None,
+    map_description: str = None,
+    is_public: bool = True,
+    colorable_fields: list = [],
+    build_topic_model: bool = False,
```
bstadt, line 25: default to tru?

nomic/atlas.py
```diff
@@ -2,1315 +2,234 @@
...
+    topic_label_field: str = None,
+    num_workers: int = 10,
+    organization_name: str = None,
+    reset_project_if_exists: bool = False,
+    add_datums_if_exists: bool = False,
+    shard_size: int = 1000,
+    projection_n_neighbors: int = DEFAULT_PROJECTION_N_NEIGHBORS,
+    projection_epochs: int = DEFAULT_PROJECTION_EPOCHS,
+    projection_spread: float = DEFAULT_PROJECTION_SPREAD,
+) -> AtlasProject:
```
bstadt, line 35: several params are not documented (projection_n_neighbors, projection_epochs, etc...)

nomic/atlas.py
```diff
@@ -2,1315 +2,234 @@
...
+            build_topic_model=build_topic_model,
+            projection_n_neighbors=projection_n_neighbors,
+            projection_epochs=projection_epochs,
+            projection_spread=projection_spread,
+            topic_label_field=topic_label_field,
+        )
+        logger.info(str(projection))
+    else:
+        # otherwise refresh the maps
+        project.refresh_maps()
```
bstadt, line 124: not sold on this auto refresh but i dont have a strong case against so prob fine for now

nomic/atlas.py
```diff
@@ -2,1315 +2,234 @@
...
+        map_name: A name for your map.
+        map_description: A description for your map.
+        build_topic_model: Builds a hierarchical topic model over your data to discover patterns.
+        multilingual: Should the map take language into account? If true, points from different with semantically similar text are considered similar.
+        is_public: Should this embedding map be public? Private maps can only be accessed by members of your organization.
+        colorable_fields: The project fields you want to be able to color by on the map. Must be a subset of the projects fields.
+        organization_name: The name of the organization to create this project under. You must be a member of the organization with appropriate permissions. If not specified, defaults to your user accounts default organization.
+        reset_project_if_exists: If the specified project exists in your organization, reset it by deleting all of its data. This means your uploaded data will not be contextualized with existing data.
+        add_datums_if_exists: If specifying an existing project and you want to add data to it, set this to true.
+
```
bstadt, line 165: also missing several args

nomic/project.py
```diff
@@ -0,0 +1,1370 @@
...
+from tqdm import tqdm
+
+import nomic
+
+from .cli import refresh_bearer_token, validate_api_http_response
+from .settings import *
+from .utils import assert_valid_project_id, get_object_size_in_bytes
+
+
+class AtlasUser:
```
bstadt, line 30: feels weird given that AtlasClass has a bearer

nomic/project.py
```diff
@@ -0,0 +1,1370 @@
...
+            response = requests.get(
+                self.atlas_api_path + "/v1/user",
+                headers=self.header,
+            )
+            response = validate_api_http_response(response)
+            if not response.status_code == 200:
+                logger.info("Your authorization token is no longer valid.")
+        else:
+            raise ValueError(
+                "Could not find an authorization token. Run `nomic login` to authorize this client with the Nomic API."
```
bstadt, line 68: python client login method?

nomic/project.py
```diff
@@ -0,0 +1,1370 @@
...
+
+        return response.json()
+
+    def _validate_map_data_inputs(self, colorable_fields, id_field, data):
+        '''Validates inputs to map data calls.'''
+
+        if not isinstance(colorable_fields, list):
+            raise ValueError("colorable_fields must be a list of fields")
+
+        if id_field in colorable_fields:
```
bstadt, line 92: hm i literally did this in my ATP map and it was fine...

bstadt: like i wanted to color progressively by it, which should be allowed methinks

nomic/project.py
```diff
@@ -0,0 +1,1370 @@
...
+        organization_name, organization_id = self._get_organization(organization_name=organization_name)
+
+        # check if this project already exists.
+        response = requests.post(
+            self.atlas_api_path + "/v1/project/search/name",
+            headers=self.header,
+            json={'organization_name': organization_name, 'project_name': project_name},
+        )
+        if response.status_code != 200:
+            raise Exception(f"Failed to create project: {response.json()}")
```
bstadt, line 307: this seems like an awk error message

nomic/project.py
```diff
@@ -0,0 +1,1370 @@
...
+        if search_results:
+            existing_project = search_results[0]
+            existing_project_id = existing_project['id']
+            return {
+                'project_id': existing_project_id,
+                'organization_id': organization_id,
+                'organization_name': organization_name,
+            }
+
+        return {'organization_id': organization_id, 'organization_name': organization_name}
```
bstadt, line 319: this return doesnt make much sense to me, and is also not documented

nomic/project.py
```diff
@@ -0,0 +1,1370 @@
...
+        * **organization_name** - The name of the organization to create this project under. You must be a member of the organization with appropriate permissions. If not specified, defaults to your user account's default organization.
+        * **is_public** - Should this project be publicly accessible for viewing (read only). If False, only members of your Nomic organization can view.
+        * **reset_project_if_exists** - If the requested project exists in your organization, will delete it and re-create it.
+        * **project_id** - An alternative way to retrieve a project is by passing the project_id directly. This only works if a project exists.
+        * **reset_project_if_exists** - If the specified project exists in your organization, reset it by deleting all of its data. This means your uploaded data will not be contextualized with existing data.
+        * **add_datums_if_exists** - If specifying an existing project and you want to add data to it, set this to true.
+        **Returns:** project_id on success.
+
+        """
+        assert name is not None or project_id is not None, "You must pass a name or project_id"
```
bstadt, line 555: why assert when you could just make them required?

nomic/project.py
```diff
@@ -0,0 +1,1370 @@
...
+
+        super().__init__()
+
+        if project_id is not None:
+            self.meta = self._get_project_by_id(project_id)
+            return
+
+
+        results = self._get_existing_project_by_name(project_name=name, organization_name=organization_name)
+        organization_name = results['organization_name']
```
bstadt, line 565: wait this is confusing so you _use_ the results of that awk return? i feel like im missing something about the behaviour here in your mental model...

nomic/project.py
```diff
@@ -0,0 +1,1370 @@
...
+        assert name is not None or project_id is not None, "You must pass a name or project_id"
+
+        super().__init__()
+
+        if project_id is not None:
+            self.meta = self._get_project_by_id(project_id)
+            return
+
+
+        results = self._get_existing_project_by_name(project_name=name, organization_name=organization_name)
```
bstadt, line 564: ah so this is why you cant return an AtlasProject...
this feels kinda awk - maybe just rename the function? like its not really getting a project.

bstadt: OH its only used in 1 place - here! just put the logic in this class?

nomic/project.py
```diff
@@ -0,0 +1,1370 @@
...
+
+        if 'project_id' in results: #project already exists
+            project_id = results['project_id']
+            if reset_project_if_exists:
+                logger.info(
+                    f"Found existing project `{name}` in organization `{organization_name}`. Clearing it of data by request."
+                )
+                self._delete_project_by_id(project_id=project_id)
+                project_id = None
+            else:
```
bstadt, line 575: elif?

nomic/project.py
```diff
@@ -0,0 +1,1370 @@
...
+                        f"Using existing project `{name}` in organization `{organization_name}`."
+                    )
+                else:
+                    raise ValueError(
+                        f"Project already exists with the name `{name}` in organization `{organization_name}`. "
+                        f"You can add datums to it by settings `add_datums_if_exists = True` or reset it by specifying `reset_project_if_exist=True` on a new upload."
+                    )
+
+
+        if project_id is None: #if there is no existing project, make a new one.
```
bstadt, line 587: can you document this in the docstring pls? like just specify when this method will create new vs use old

nomic/project.py
```diff
@@ -0,0 +1,1370 @@
...
+
+            if unique_id_field is None:
```
bstadt, line 589: why? dont we have an internal one? and if you have to specify, why is it default None?

nomic/project.py
```diff
@@ -0,0 +1,1370 @@
...
+            reuse_embeddings_from_index: the name of the index to reuse embeddings from.
+
+        Returns:
+            The projection this index has built.
+
+        '''
+
+        self._latest_project_state()
+
+        # for large projects, alter the default projection configurations.
```
bstadt, line 837: nice

bstadt (approved):

nomic/project.py
AndriyMulyar: because one may want to instantiate the class by using a project_id for example and they should be only able to supply that.

PR #74, opened 2023-02-13
Support and documentation for dynamic maps, simple tagging tutorial, bug fixes

AndriyMulyar:


examples/map_embeddings_progressively.py
```diff
@@ -1,22 +1,24 @@
...
-response = atlas.update_maps(embeddings=second_upload_embeddings,
-                             data=second_upload_data)
+# embeddings with shifted mean.
+embeddings += np.ones(shape=(num_embeddings, 10))
+data = [{'upload': '2'} for i in range(len(embeddings))]
 
-print('Second upload response: ', response)
+with project.block_until_accepting_data() as project:
+    project.add_embeddings(embeddings=embeddings, data=data)
+    project.rebuild_maps()
```
bmschmidt, line 24: I wonder if it would be clearer to have the block method be called `project.wait_for_upload_lock` or something?
`block_until_accepting_data` to me implies that the context manager *creates* a block, rather than waits for an existing one to be released.

nomic/project.py
```diff
@@ -755,6 +756,17 @@ def is_accepting_data(self) -> bool:
         self._latest_project_state()
         return not self.is_locked
 
+
+    @contextmanager
+    def block_until_accepting_data(self):
```
bmschmidt, line 761: See above.

nomic/project.py
```diff
@@ -963,11 +967,62 @@ def get_projection_id():
 
     def __repr__(self):
         m = self.meta
-        return f"Nomic project: <{m}>"
+        return f"AtlasProject: <{m}>"
```
bmschmidt, line 970: I will add a _repr_html_ method for Ipython.

nomic/project.py
```diff
@@ -963,11 +967,62 @@ def get_projection_id():
...
+            raise Exception(response.json())
+
+    def delete_data(self, ids: List[str]) -> bool:
+        '''
+        Deletes the specified datums from the project.
+
+        Args:
+            ids: A list of datum ids to delete
+
+        Returns:
```
bmschmidt, line 1009: This returns `True`. Perhaps it should return `None` on successful completion--that's more the Unix standard.

nomic/project.py
```diff
@@ -1336,8 +1392,6 @@ def refresh_maps(self):
 
         logger.info(f"Updating maps in project `{self.name}`")
 
-        return self
-
     def get_tags(self) -> Dict[str, str]:
```
bmschmidt, line 1395: Type signature is Dict[str, str], but function returns `Dict[str, List[str]]`.

I find it hard to decide if the keys of this dictionary should be the tag names, or the datum ids. Seems like the code here also goes back and forth.

We should think about how to support a workflow that wants to get all the tags associated with an id, not just all the ids associated with a tag. 

TBH, my preferred format would be to return either a list of dicts ([{id, tag}]) or a list of two-tuples ([(id, tag)]).

nomic/project.py
```diff
@@ -963,11 +967,62 @@ def get_projection_id():
...
     def __repr__(self):
         m = self.meta
-        return f"Nomic project: <{m}>"
+        return f"AtlasProject: <{m}>"
 
     def __str__(self):
         return "\n".join([str(projection) for index in self.indices for projection in index.projections])
 
+
+    def get_data(self, ids: List[str]) -> List[Dict]:
```
bmschmidt, line 976: Why shouldn't get_data include user-generated tags as part of the return signature? Isn't that part of the dataset?

docs/data_cleaning_in_atlas.md
```diff
@@ -0,0 +1,99 @@
...
+
+Next, use the lasso tool to tag regions of interest. In this example, we're going to remove
+all the sports related news articles from the dataset.
+
+Notice, that all the sports articles are pre-organized into one region of the map. This is because the neural
+networks that power Atlas produce vectors for sports articles that are all similar to each other.
+
+![](assets/labeling_tutorial/ag_news_10k_news_lassoed.png)
+
+After tagging this lassoed region as 'sports', we can access all the tagged points programmatically.
```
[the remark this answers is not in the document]
AndriyMulyar: This will be modified once we have the workflow in the UI set in stone

nomic/project.py
AndriyMulyar: Can we somehow detect if a user is in a jupyter notebook and just do the right thing instead?

nomic/project.py
AndriyMulyar: The endpoint from atlas returns a mapping from ids to tags (the first thing you said). I think we should overload this functions parameters to specify if you want id -> tag or tag -> id

But I agree having both this good.

AndriyMulyar: Made a card.

examples/map_embeddings_progressively.py
AndriyMulyar: Didn't really think of it that way. It isnt just an upload block, it also blocks operations such a reindexing, nn search, etc.

I would be open to changing it to wait_for_lock_release() or something akin.

Thoughts?

docs/data_cleaning_in_atlas.md
```diff
@@ -0,0 +1,99 @@
...
+    project = atlas.map_text(data=documents,
+                             id_field='id',
+                             indexed_field='text',
+                             map_name='News 10k Labeling Example',
+                             map_description='10k News Articles for Labeling'
+                             )
+    print(project.maps)
+    ```
+
+Produces a map that should roughly resemble this:
```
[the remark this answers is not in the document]
AndriyMulyar: We're roughly deterministic lol. Initialization puts us in the right ball park but clusters will be at *slightly* different positions on different uploads. Not sure exactly why likely non randomness in the projection. 
But I will remove the word.

nomic/project.py
AndriyMulyar: made a card

nomic/project.py
AndriyMulyar: https://stackoverflow.com/questions/15411967/how-can-i-check-if-code-is-executed-in-the-ipython-notebook

we can

PR #81, opened 2023-02-16
Vector search error logging

AndriyMulyar:


bstadt (approved):

nomic/project.py
```diff
@@ -513,32 +513,31 @@ def vector_search(self, queries: np.array = None, ids: List[str] = None, k: int
...
+                      'datum_ids': ids,
+                      'k': k},
+            )
 
-            status = response.status_code
-            retries -= 1
+        status = response.status_code
 
-        if retries == 0:
+        if response.status_code == 500:
```
bstadt, line 535: would love some logging here as well

AndriyMulyar: It throws an exception, log message would be redundant.

PR #83, opened 2023-02-17
just name/description, not project_name

bmschmidt:
YMMV, but I find it confusing to have this argument called `map_name` that turns into `project_name` and `index_name`. Plus the arguments to the `AtlasProject` constructor aren't `project_name` and `project_description`, they're just `name` and `description`. I propose that we do the same thing for `map_embeddings` and `map_text`--makes it easier to switch between the quickstart workflow, avoids adding `map_name` as a nonexistent identifier into our implicit namespace.

### DIFF
--- a/examples/interactive_session.py
+++ b/examples/interactive_session.py
@@ -10,9 +10,9 @@ documents = [dataset[i] for i in subset_idxs]
 
 project = atlas.map_text(data=documents,
                           indexed_field='text',
-                          map_name='News Dataset 25k',
+                          name='News Dataset 25k',
                           colorable_fields=['label'],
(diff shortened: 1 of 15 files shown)


AndriyMulyar:
Ignore this failing commit, i blundered

examples/interactive_session.py
```diff
@@ -10,9 +10,9 @@
 
 project = atlas.map_text(data=documents,
```
AndriyMulyar, line 11: Okay, i'm fine with this. Just make sure to change all of the documentation

AndriyMulyar (approved):

AndriyMulyar (approved):

PR #85, opened 2023-02-17
Ipyframes

bmschmidt:
This adds `_repr_html_` to AtlasProject, AtlasProjection, and AtlasIndex classes that includes, frequently, and embedded iframe to the Atlas viewer of the project.

If there's a complete projection build in the project (determined through calling the API projection status endpoint)
it adds the most recent one to the `AtlasProject` display, and it shows it for AtlasProjection. Otherwise, it just gives the status and a link to Atlas.

There's a place to delete the big interactive memory hog if you don't like it.

### DIFF
--- a/nomic/project.py
+++ b/nomic/project.py
@@ -346,7 +346,7 @@ class AtlasProjection:
     Instead instantiate an AtlasProject and use the project.indices or get_map method to retrieve an AtlasProjection.
     '''
 
-    def __init__(self, project : AtlasProject, atlas_index_id: str, projection_id: str, name):
+    def __init__(self, project : "AtlasProject", atlas_index_id: str, projection_id: str, name):
         """
         Creates an AtlasProjection.
         """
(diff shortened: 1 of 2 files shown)


nomic/project.py
```diff
@@ -344,7 +346,8 @@ class AtlasProjection:
     Instead instantiate an AtlasProject and use the project.indices or get_map method to retrieve an AtlasProjection.
     '''
 
-    def __init__(self, project, atlas_index_id: str, projection_id: str, name):
+
+    def __init__(self, project : "AtlasProject", atlas_index_id: str, projection_id: str, name):
```
AndriyMulyar, line 350: Ahh this is how you type it

AndriyMulyar (approved):
Lgtm

nomic/project.py
bmschmidt: Only in 3.7 and below! Works fine later.

PR #110, opened 2023-03-23
Updated documentation around timestamps

AndriyMulyar:


docs/mapping_faq.md
```diff
@@ -48,21 +48,11 @@ If you are added to a Nomic organization by someone (such as your employer), you
...
-```
-If you are working with dates which are in a non-uniform format, parsing into datetime objects may be difficult. Nomic recommends
-using the `python-dateutil` package in this situation. It will intelligently parse a string into a Python datetime object at the cost of some compute cycles.
-```python
-from dateutil import parser
-date = parser.parse("Apr 15 1999 12:00AM")  # datetime.datetime(1999, 4, 15, 0, 0)
-date.isoformat()
-``` 
+## Working with Dates and Timestamps
+Atlas will consider metadata as timestamps when they are processed into Python `datetime` objects. Under the hood,
```
bmschmidt, line 52: "....when they are passed as Python `date` or `datetime` objects."

bmschmidt (approved):

bmschmidt (commented):
Great.

PR #137, opened 2023-04-14
Embedding iterator fix

bmschmidt:
@zanussbaum needs this now.

### DIFF
--- a/nomic/project.py
+++ b/nomic/project.py
@@ -558,15 +558,15 @@ class AtlasProjection:
 
         '''
 
-        if self.is_locked:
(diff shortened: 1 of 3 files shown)


nomic/project.py
```diff
@@ -548,15 +558,15 @@ def get_embedding_iterator(self) -> Iterable[Tuple[str, str]]:
 
         '''
 
-        if self.is_locked:
-            raise Exception('Project is locked! Please wait until the project is unlocked to download embeddings')
+        #if False or self.is_locked:
```
AndriyMulyar, line 561: this should be self.project.is_locked

AndriyMulyar (changes requested):

PR #158, opened 2023-05-19
[RFR] reserve dollar sign fields

bmschmidt:
I want to place a hold on all field names that begin with a dollar sign. I'm preliminarily thinking of using this for some kind of function application or selection reference application inside deepscatter--even if we don't do that, I think it might be useful and not disruptive to block off this namespace for ourselves.


bmschmidt:
ship has sailed.

PR #166, opened 2023-06-05
allow overriding api path with env var

apage43:
this allows testing against a local atlas dev env (once `nomic login`d as a staging user) with e.g.

```bash
ATLAS_API_PATH=http://localhost make test
```


bmschmidt (approved):
Looks good, and like the path of least resistance here.

PR #167, opened 2023-06-05
save arrow format embeddings as-is

apage43:
only land once that's what's actually served here


nomic/project.py
```diff
@@ -516,12 +519,12 @@ def download_shard(offset, check_access=False):
...
             try:
-                content = response.json()
+                content = response.content
 
-                shard_name = '{}_{}_{}.pkl'.format(self.atlas_index_id, offset, offset + limit)
+                shard_name = '{}_{}_{}.feather'.format(self.atlas_index_id, offset, offset + limit)
                 shard_path = os.path.join(save_directory, shard_name)
                 with open(shard_path, 'wb') as f:
-                    pickle.dump(content, f)
+                    f.write(content)
```
bmschmidt, line 527: Should we validate by reading the bytes into Arrow? If not might be worth checking that:
```
content[:6] == b"ARROW1" and content[-6:] == b"ARROW1"
```

bmschmidt: (But maybe we have a status==200 check above that guarantees this?)

apage43: wouldn't hurt to be sure the format is right and not just that the status is correct, especially since the API changed

AndriyMulyar:
ci should be fixed now (was screwed up earlier)

apage43:
CI didn't trigger this time either?

AndriyMulyar (approved):

PR #170, opened 2023-06-08
Download sidecars along with feathers

bmschmidt:
Untested as of yet, but maybe useful for @rguo123 soon.


bmschmidt:
Now used with an ipython notebook. Change the project id to your own to run.

PR #177, opened 2023-06-09
v2.0.0: Implemented and moved all examples, documentation and tests over to namespaces data access attributes

AndriyMulyar:
# v2.0.0

Atlas map state is now accessible as top level attributes of the AtlasProjection ("the Map") class.
```python
from atlas import AtlasProject

project = AtlasProject(name='My Project')
map = project.maps[0]

```

- Topics (`map.topics`)
- Embeddings (`map.embeddings`)
- Semantic Duplicate Clusters (`map.duplicates`)

You can learn how to use the operations these attributes encapsulate on their corresponding data in the `Accessing Atlas State` section of the documentation.

------
Updated documentation and example scripts to reflect this change.


docs/map_state/embeddings.md
```diff
@@ -0,0 +1,24 @@
...
+
+
+```python
+from nomic import AtlasProject
+
+project = AtlasProject(name='My Project')
+
+map = project.maps[0]
+
+map.embeddings.df
```
zanussbaum, line 14: My first natural question here is what columns can we expect? Will we get both the ambient and 2d representations in the same dataframe?

docs/map_state/tags.md
```diff
@@ -0,0 +1,26 @@
...
+
+
+```python
+from nomic import AtlasProject
+
+project = AtlasProject(name='My Project')
+
+map = project.maps[0]
+
+map.tags.get_tags()
```
zanussbaum, line 16: Should we note here that tags returns a dictionary where the keys are the tag names the user adds in the frontend and the values are a list of ids? I found this confusing when I was initially using `get_tags`.

Should we also mention that if the use doesn't provide `id_field` in `map_text`, it's very tough to map from tags to your original dataframe? cc @bmschmidt

docs/map_state/topics.md
```diff
@@ -0,0 +1,25 @@
...
+
+
+```python
+from nomic import AtlasProject
+
+project = AtlasProject(name='My Project')
+
+map = project.maps[0]
+
+map.topics.df
```
zanussbaum, line 15: nit: same thing here, can we either show an example or explain what the dataframe looks like (e.g. what columns)

nomic/atlas.py
```diff
@@ -76,9 +78,7 @@ def map_embeddings(
         description = description
 
     if data is None:
-        data = [{
-            ATLAS_DEFAULT_ID_FIELD: b64int(i)
-        } for i in range(len(embeddings))]
+        data = [{ATLAS_DEFAULT_ID_FIELD: b64int(i)} for i in range(len(embeddings))]
```
zanussbaum, line 81: it would be great if we either enforce users to upload with an id field or find a way to pass that to the user so they can map back from `datum_id` to their original data. I've had to rebuild maps a few too many times since I forgot to add an id column and couldn't join with the tags I had added to the data points

nomic/data_operations.py
```diff
@@ -1,42 +1,602 @@
...
+        Two-dimensional embeddings.
+
+        These are the points you see in your web browser.
+
+        Returns:
+            Pandas dataframe mapping your datapoints to their two-dimensional embeddings.
+        """
+        return self.df
+
+    # @property
```
zanussbaum, line 321: Should we update the documentation if this isn't finished yet?

nomic/data_operations.py
```diff
@@ -1,42 +1,602 @@
...
+        writer = ipc.new_file(buffer, payload_table.schema, options=ipc.IpcWriteOptions(compression='zstd'))
+        writer.write_table(payload_table)
+        writer.close()
+        payload = buffer.getvalue()
+
+        headers = self.project.header.copy()
+        headers['Content-Type'] = 'application/octet-stream'
+        response = requests.post(self.project.atlas_api_path + "/v1/project/tag/delete", headers=headers, data=payload)
+        if response.status_code != 200:
+            raise Exception("Failed to delete tags")
```
zanussbaum, line 602: do we want a `__repr__` method to stay consistent with the other classes?

nomic/data_operations.py
```diff
@@ -1,42 +1,602 @@
...
+            content = response.json()
+            if len(content['datum_ids']) == 0:
+                break
+            offset += len(content['datum_ids'])
+
+            yield content['datum_ids'], content['embeddings']
+
+    def download_embeddings(self, save_directory: str, num_workers: int = 10) -> bool:
+        '''
+        Downloads shards of arrow tables that map
```
zanussbaum, line 433: nit: is this complete?

nomic/gpt4all/gpt4all.py
```diff
@@ -23,19 +26,18 @@ def __init__(self, llama_path=None):
         self.model_path = llama_path
         self.tokenizer_path = llama_path
         self.lora_path = 'nomic-ai/vicuna-lora-multi-turn_epoch_2'
```
zanussbaum, line 28: wait is the gpt4all model the first model we trained? should we update this to a newer one?

zanussbaum (commented):
Had a few comments, mostly on documentation + clarifications

nomic/atlas.py
AndriyMulyar: yeah, they just shouldn't be using map_embeddings or map_text.
its kinda a hard problem to solve.

nomic/data_operations.py
AndriyMulyar: What should this be. The issue with tags is that we don't get global access of all tag state

nomic/data_operations.py
AndriyMulyar: all of AtlasMapEmbeddings is not fully complete yet. it is blocked by ben implementing logic on the atlas cloud backend for writing embeddings more efficiently for export.

Once we have an efficient method pull out ambient embeddings, accessing with .ambient will be possible.

For now the existing logic where we download all embeddings to a directory is here.

nomic/gpt4all/gpt4all.py
AndriyMulyar: We should remove gpt4all from the repository, but lets do it in a separate PR

nomic/atlas.py
AndriyMulyar: I will add information that these are set in upload order.

nomic/data_operations.py
AndriyMulyar: I will add a .df method for consistency

docs/data_exploration_cleaning_tagging_in_atlas.ipynb
```diff
@@ -172,7 +172,7 @@
     }
    ],
    "source": [
-    "tags = map.get_tags()\n",
+    "tags = map.tags.get_tags()\n",
```
bmschmidt, line 175: Ideally `map.tags` would do the retrieval on its own, not requiring a call to `map.tags.get_tags()`.

docs/map_state/tags.md
bmschmidt: We should make sure all our examples show people uploading ids.

With the new auto assignment it's harder to screw it up... but still not great.

docs/map_state/topics.md
```diff
@@ -0,0 +1,25 @@
+Atlas pre-organizes your data into topics informed by the latent contents of your embeddings. Visually, these are represented by regions of homogenous
```
bmschmidt, line 1: 'regions of homogenous color' -> 'regions of the same color'

docs/release_notes.md
```diff
@@ -1,6 +1,24 @@
...
+map = project.maps[0]
+
+```
+
+- Topics (`map.topics`)
+- Embeddings (`map.embeddings`)
+- Semantic Duplicate Clusters (`map.duplicates`)
+- Tagging (`map.tags`)
+
+You can learn how to use the operations these attributes encapsulate on their corresponding data in the `Accessing Atlas State` section of the documentation.
```
bmschmidt, line 20: Better--"The `Accessing Atlas State` section of the documentation shows some of the ways you can use this data."

nomic/atlas.py
bmschmidt: I think we could write a warning to stderr on map_embeddings and map_text.

nomic/gpt4all/gpt4all.py
bmschmidt: Removal should definitely be part of the 2.0 release.

nomic/project.py
```diff
@@ -8,46 +8,42 @@
...
 
-
 # Mainly for type checking
 try:
-    from pandas import DataFrame
     import pandas as pd
+    from pandas import DataFrame
 except ImportError:
     pd = None
     DataFrame = None
```
bmschmidt, line 31: We can lose this try-catch; pandas is a dependency again.

nomic/project.py
```diff
@@ -366,14 +365,15 @@ def __init__(self, atlas_index_id, name, indexed_field, projections):
     def _repr_html_(self):
         return '<br>'.join([d._repr_html_() for d in self.projections])
 
+
 class AtlasProjection:
     '''
-    Manages operations on maps such as text/vector search.
+    Interact and access state of an Atlas Map including text/vector search.
     This class should not be instantiated directly.
     Instead instantiate an AtlasProject and use the project.indices or get_map method to retrieve an AtlasProjection.
```
bmschmidt, line 373: Instead instantiate an AtlasProject and use the project.indices or get_map method to retrieve an AtlasProjection.
should be
    Instead instantiate an AtlasProject and use the project.maps attribute to retrieve an AtlasProjection.

nomic/project.py
```diff
@@ -1370,7 +1006,9 @@ def create_index(
                         reuse_embedding_from_index_id = index.id
                         break
                 if reuse_embedding_from_index_id is None:
-                    raise Exception(f"Could not find the index '{reuse_embeddings_from_index}' to re-use from. Possible options are {[index.name for index in indices]}")
+                    raise Exception(
+                        f"Could not find the index '{reuse_embeddings_from_index}' to re-use from. Possible options are {[index.name for index in indices]}"
+                    )
```
bmschmidt, line 1011: Seeing how aggressively we use the word 'Index' here, I'm wondering if we should make atlas.maps return all the indices, rather than the projections.... I think it will be confusing for people to encounter the idea of an index if we're calling them 'maps'

docs/map_state/topics.md
AndriyMulyar: Resolved by adding examples containing the dataframe structure

nomic/atlas.py
AndriyMulyar: I will add a warning and also documentation.

nomic/gpt4all/gpt4all.py
AndriyMulyar: 🪓 done

nomic/data_operations.py
AndriyMulyar: I have added a .df method which contains as columns a multi-hot vector where ones exist when the associated row has that tag.

docs/data_exploration_cleaning_tagging_in_atlas.ipynb
AndriyMulyar: Done

docs/map_state/duplicates.md
```diff
@@ -0,0 +1,26 @@
+Atlas groups your data into semantically similar clusters powered by latent information contained in your embeddings.
+Under the hood, it utilizes an algorithm similar to [SemDeDup](https://arxiv.org/abs/2303.09540).
+
+
+You can access and operate on semantic duplicate clusters programmatically by using the `duplicates` attribute
+of an AtlasMap. 
+
```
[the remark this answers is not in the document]
AndriyMulyar: I think this would make a lot of sense.

docs/map_state/embeddings.md
```diff
@@ -0,0 +1,24 @@
+Atlas stores, manages and generates embeddings for your unstructured data.
+
+You can access Atlas embeddings in their ambient form (e.g. high dimensional) or in their two-dimensional
+projected representations.
```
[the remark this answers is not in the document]
AndriyMulyar: embedded sounds confusing (map.embeddings.embedded?)

i am fine with either latent/ambient.


Another idea: what if we called them
two_dimensional
high_dimensional

instead?

nomic/atlas.py
zanussbaum: What should people be using instead of map_embeddings or map_text then?

nomic/project.py
AndriyMulyar: It is certainly confusing to have many abstraction, we have talked in person about this in-depth before.

Currently AtlasIndex is a proxy container from the ages where we were not certain what Atlas would be so built in room for flexibility.

I would propose we just *ignore* it and only mention AtlasProject and AtlasProjection ("map") across all documentation. We could essentially replace the word index to map in every location and no-one would be the wiser (and the code may be the clearer) as AtlasIndex is currently a passthrough object.

nomic/data_operations.py
zanussbaum: yeah good point, resolving

AndriyMulyar:
- Removed references to GPT4All
- Added warning when people don't specify ids
- Gave `map.tags` a dataframe that can be sliced and diced (but it is materialized from the server state on every call)
- Added more code/output examples for all data state classes

nomic/data_operations.py
```diff
@@ -24,6 +24,27 @@
 class AtlasMapDuplicates:
     """
     Atlas Duplicate Clusters State
+
```
zanussbaum, line 27: i know these were probably a pain to write up but this is very much appreciated and really improves on the docs 😄

docs/map_state/embeddings.md
```diff
@@ -7,18 +7,12 @@ projected representations.
 ```python
 from nomic import AtlasProject
 
-project = AtlasProject(name='My Project')
+map = AtlasProject(name='My Project').maps[0]
```
zanussbaum, line 10: Probably not the place for this, but most of my workflow has been creating a single map per project. What's the use case for multiple maps per single project?

nomic/atlas.py
```diff
@@ -77,13 +77,19 @@ def map_embeddings(
...
         data = [{ATLAS_DEFAULT_ID_FIELD: b64int(i)} for i in range(len(embeddings))]
+        added_id_field = True
 
     if id_field == ATLAS_DEFAULT_ID_FIELD and id_field not in data[0]:
+        added_id_field = True
         for i in range(len(data)):
             data[i][id_field] = b64int(i)
 
+    if added_id_field:
+        logger.warning("An ID field was not specified in your data so one was generated for you..")
```
zanussbaum, line 91: Should we mention the ID is in order from `range(len(data))` in order that the data was added?

zanussbaum (approved):
LGTM pending Ben's review

nomic/atlas.py
AndriyMulyar: p = AtlasProject()

p.add_text()

p.create_index()

nomic/data_operations.py
AndriyMulyar: No, this is what the world needed

docs/map_state/embeddings.md
AndriyMulyar: if you index the project multiple times (say on different fields), you will have a map per index so possibly multiple indices

nomic/atlas.py
AndriyMulyar: updated

docs/map_state/embeddings.md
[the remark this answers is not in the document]
bmschmidt: I'm lukewarm on `two_dimensional`/`high_dimensional`. Partly because I don't like writing out numbers.

My thought on 'embedded' is that accessing embeddings does just by default work on the latent space, and the two-d space is a special case. It's not like 'embeddings.vector_search' lets you search in the 2d space as an option.

So maybe like `map.embeddings.array`, `map.embeddings.ndarray`, `map.embeddings.matrix` just returns the latent space as an `n x d` np.ndarray, and `map.embeddings.projected` returns the `n x 2` array.

nomic/project.py
bmschmidt: That sounds good to me--so this method will be renamed `reuse_embeddings_from_map`?

docs/data_exploration_cleaning_tagging_in_atlas.ipynb
rguo123: Notebook needs to be updated as well?

docs/map_state/duplicates.md
```diff
@@ -0,0 +1,20 @@
...
+
+```python
+from nomic import AtlasProject
+
+map = AtlasProject(name='My Project').maps[0]
```
rguo123, line 12: super nitpick: In one of the tutorials, it may be good to showcase `get_map` instead of accessing the list of maps?

docs/map_state/topics.md
```diff
@@ -0,0 +1,20 @@
...
+color on an Atlas map.
```
rguo123, line 2: Nit: "AtlasMap" since that's used elsewhere

docs/vector_database.md
```diff
@@ -6,7 +6,10 @@ This tutorial will show you how you can visualize your Weaviate and Pinecone vec
 
 ## Why Visualize A Vector Database
 
-Vector databases allow you to query your data semantically by indexing embedding vectors. By interactively visualizing embeddings, you can quickly understand the space of possible query results from your vector database and find bad embeddings which may produce poor query results.
+Vector databases allow you to query your data semantically by indexing embedding vectors. By interactively visualizing embeddings, you can quickly:
+
+- Understand the space of possible query results from your vector database
+- Identify bad embeddings in your index which may produce poor query results.
```
rguo123, line 12: Nit: no period

nomic/pl_callbacks/pl_callback.py
```diff
@@ -45,8 +52,17 @@ def clear(self):
         self.embeddings = []
         self.metadata = defaultdict(list)
 
+
 class AtlasEmbeddingExplorer(Callback):
-    def __init__(self, max_points=-1, rebuild_time_delay=600, name=None, description=None, is_public=True, overwrite_on_validation=False):
+    def __init__(
+        self,
+        max_points=-1,
```
rguo123, line 59: As part of docs, can we include what -1 is referring to here?

AndriyMulyar: done

PR #179, opened 2023-06-20
Duplicate parameter

bmschmidt:
For review, not yet merge.

Sets up a pattern for subclasses accessing different created metadata in Atlas.

Here, only duplicates implemented so we can think through the patterns.

See the included IPython notebook for an example.

### DIFF
--- a/nomic/data_operations.py
+++ b/nomic/data_operations.py
@@ -1,6 +1,42 @@
 import pandas as pd
-
-class AtlasDuplicates(pd.DataFrame):
-    def __init__(self, projection : AtlasProjection):
+import pyarrow as pa
+from pyarrow import compute as pc
+from typing import List
+class AtlasDuplicates():
+    """
+    The duplicates 
+    """
+    def __init__(self, projection : "AtlasProjection"):
         self.projection = projection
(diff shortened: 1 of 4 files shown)


nomic/data_operations.py
```diff
@@ -0,0 +1,42 @@
...
+import pyarrow as pa
+from pyarrow import compute as pc
+from typing import List
+class AtlasDuplicates():
+    """
+    The duplicates 
+    """
+    def __init__(self, projection : "AtlasProjection"):
+        self.projection = projection
+        self.id_field = self.projection.project.id_field
```
AndriyMulyar, line 11: very cool that its this simple. we should brainstorm what operations for topics, duplicates, etc we want to include as defacto methods

AndriyMulyar (approved):

AndriyMulyar:
the failing test here is fine, its because we are missing the embedding downloads updates in prod. should go out today.

AndriyMulyar:
merge from main and then re-run unit tests, @apage43 updates to embedding download was just fixed

AndriyMulyar (approved):

PR #194, opened 2023-07-06
Embedding import

bmschmidt:
Adds an nparray at `Map.embeddings.latent`.

Notes:
* Uses single-threaded download: on contrary wifi, this takes 4 min 19 seconds to download a 1649399 dataset. Could be significantly faster.
* Returns embeddings as `float16`, which some people might find weird.
* Provides a consistent traversal order over data.
* Does not have a tqdm progress bar for the download'
* Wrote a new download method and deprecation-error the old one rather than just overwrite it.
* Does not yet include tests.

### DIFF
--- a/nomic/data_operations.py
+++ b/nomic/data_operations.py
@@ -7,11 +7,12 @@ import os
 from collections import defaultdict
 from datetime import datetime
 from typing import Dict, Iterable, List, Tuple
-
+from io import BytesIO
 import numpy as np
 import pandas
 import pandas as pd
 import pyarrow as pa
+from pyarrow import feather
 import requests
 from loguru import logger
 from pyarrow import compute as pc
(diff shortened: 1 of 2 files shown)


nomic/data_operations.py
```diff
@@ -409,15 +411,59 @@ def projected(self) -> pd.DataFrame:
...
-        # 1. download embeddings and store it in a fixed location on disk (e.g. .nomic directory)
-        # 2. make sure the embeddings align with the arrow table download order.
-        # """
-        raise NotImplementedError(
-            "Accessing latent embeddings is not yet implemented. You must use map.download_embeddings() method for now."
-        )
+    def latent(self) -> np.array:
+        """
+        Returns the embeddings in the high-dimensional space. This returns a memmapped 
+        numpy array where each row corresponds to the datapoint
```
bmschmidt, line 417: Finish docstring

AndriyMulyar (changes requested):
Unit tests please

PR #205, opened 2023-08-24
[RFR] Metadata download

rguo123:


nomic/data_operations.py
```diff
@@ -846,3 +846,90 @@ def remove(self, ids: List[str], tags: List[str], delete_all: bool = False) -> b
...
+                quad_str = "/".join([str(q) for q in quad])
+                encoded_colname = base64.urlsafe_b64encode(sidecar.encode('utf-8')).decode("utf-8")
+                filename = quad_str + "." + encoded_colname + ".feather"
+                path = self.projection.tile_destination / filename
+                quad_files.append(path)
+            
+                data = requests.get(root + filename)
+                readable = io.BytesIO(data.content)
+                readable.seek(0)
+                tb = feather.read_table(readable)
```
bmschmidt, line 915: Probably want to add `memory_map = True` here.

nomic/data_operations.py
```diff
@@ -846,3 +846,90 @@ def remove(self, ids: List[str], tags: List[str], delete_all: bool = False) -> b
...
+            small_sidecars = []
+        for path in self.projection._tiles_in_order():
+            tb = pa.feather.read_table(path).select(["_id"])
+            for sidecar_file in small_sidecars:
+                carfile = pa.feather.read_table(path.parent / f"{path.stem}.{sidecar_file}.feather")
+                for col in carfile.column_names:
+                    tb = tb.append_column(col, carfile[col])
+            for big_sidecar in additional_sidecars:
+                fname = base64.urlsafe_b64encode(big_sidecar.encode('utf-8')).decode("utf-8")
+                carfile = pa.feather.read_table(path.parent / f"{path.stem}.{fname}.feather")
```
bmschmidt, line 884: Probably want to add `memory_map = True` here. These files will get *big*.

bmschmidt (commented):
Looks good so far.

nomic/data_operations.py
rguo123: Added. 
Question: all the data contents for each feather file are sent over an http request and read into a pyarrow table as well for writing to disk. Are there concerns there as well?

bmschmidt: They should be one at a time, which hopefully won't be an issue. The issue is if we read them all at once.

nomic/data_operations.py
```diff
@@ -846,3 +846,125 @@ def remove(self, ids: List[str], tags: List[str], delete_all: bool = False) -> b
...
+
+    def _fetch_tiles_with_all_sidecars(self, additional_sidecars=None):
+        tbs = []
+        root = feather.read_table(self.projection.tile_destination / "0/0/0.feather")
+        try:
+            small_sidecars = set(
+                [v for k, v in json.loads(root.schema.metadata[b"sidecars"]).items()]
+            )
+        except KeyError:
+            small_sidecars = []
```
bmschmidt, line 896: For type consistency, `small_sidecars = set([])`.

nomic/data_operations.py
```diff
@@ -846,3 +846,125 @@ def remove(self, ids: List[str], tags: List[str], delete_all: bool = False) -> b
...
+        ```
+    """
+
+    def __init__(self, projection: "AtlasProjection"):
+        self.projection = projection
+        self.project = projection.project
+        self.id_field = self.projection.project.id_field
+        try:
+            # Run fetch_tiles first to guarantee existence of quad feather files
+            self._tb: pa.Table = self.projection._fetch_tiles()
```
bmschmidt, line 881: I'm having a little trouble following this, but IMO we should do here 
```
self._tb = None # Reserve this name for the actual table of all the data returned here.
try:
    self._basic_data = self.projection._fetch_tiles()
# Maybe drop the topics because they're different
```
b/c 1. that returns an arrow table.

nomic/data_operations.py
```diff
@@ -846,3 +846,125 @@ def remove(self, ids: List[str], tags: List[str], delete_all: bool = False) -> b
...
+            for sidecar in sidecars:
+                quad_str = "/".join([str(q) for q in quad])
+                encoded_colname = base64.urlsafe_b64encode(
+                    sidecar.encode("utf-8")
+                ).decode("utf-8")
+                filename = quad_str + "." + encoded_colname + ".feather"
+                path = self.projection.tile_destination / filename
+                quad_files.append(path)
+
+                # WARNING: Potentially large data request here
```
bmschmidt, line 946: I'm not super-worried about this--it's at most 50,000 points times 100kb per point, which is... hmm like a few gigabytes. Yeah that's pretty big actually. Might want to switch to just writing the bytes as they come in. https://stackoverflow.com/questions/16694907/download-large-file-in-python-with-requests

nomic/data_operations.py
```diff
@@ -846,3 +846,125 @@ def remove(self, ids: List[str], tags: List[str], delete_all: bool = False) -> b
...
+                data = requests.get(root + filename)
+                readable = io.BytesIO(data.content)
+                readable.seek(0)
+                tb = feather.read_table(readable)
+                path.parent.mkdir(parents=True, exist_ok=True)
+                feather.write_feather(tb, path)
+            returned_files.append(quad_files)
+        return sidecars, returned_files
```
bmschmidt, line 954: I think the return signature of this function could be a little cleaner. I think just a dict of `{sidecar_name: encoded_colname}`

nomic/data_operations.py
```diff
@@ -846,3 +846,125 @@ def remove(self, ids: List[str], tags: List[str], delete_all: bool = False) -> b
...
+        try:
+            # Run fetch_tiles first to guarantee existence of quad feather files
+            self._tb: pa.Table = self.projection._fetch_tiles()
+            sidecars, _ = self._download_data()
+            self._fetch_tiles_with_all_sidecars(sidecars)
+
+        except pa.lib.ArrowInvalid as e:
+            raise ValueError("Failed to fetch tiles for this map")
+
+    def _fetch_tiles_with_all_sidecars(self, additional_sidecars=None):
```
bmschmidt, line 888: This method shouldn't be called 'fetch', it should be called `_read_prefetched_tiles_with_sidecars()`

nomic/data_operations.py
bmschmidt: Or even just sidecars, looks like you don't use the encoded_colname.

nomic/data_operations.py
```diff
@@ -846,3 +846,125 @@ def remove(self, ids: List[str], tags: List[str], delete_all: bool = False) -> b
...
+                tb = feather.read_table(readable)
+                path.parent.mkdir(parents=True, exist_ok=True)
+                feather.write_feather(tb, path)
+            returned_files.append(quad_files)
+        return sidecars, returned_files
+
+    @property
+    def df(self) -> pandas.DataFrame:
+        """
+        A pandas dataframe associating each datapoint on your map to their metadata.
```
bmschmidt, line 959: We may need to add a warning on this method... after all the work we've done to avoid materializing in memory, this may materialize a lot of it.

Or eventually an option to use only specific columns.... probably not necessary immediately though.

nomic/project.py
```diff
@@ -556,7 +564,10 @@ def children(z, x, y):
             z, x, y = paths.pop(0)
             path = Path(self.tile_destination, str(z), str(x), str(y)).with_suffix(".feather")
             if path.exists():
-                yield path
+                if coords_only:
+                    yield (z, x, y)
+                else:
+                    yield path
```
bmschmidt, line 570: Good

nomic/tests/test_atlas_client.py
```diff
@@ -188,6 +188,31 @@ def test_topics():
 
         project.delete()
 
+def test_data():
+    num_embeddings = 100
+    embeddings = np.random.rand(num_embeddings, 10)
+    texts = ['foo', 'bar', 'baz', 'bat']
+    data = [
+        {'field': str(uuid.uuid4()), 'id': str(uuid.uuid4()), 'upload': 0.0, 'text': texts[i % 4]}
```
bmschmidt, line 196: Could we change this to `'text': str(i)` for this test? Once we remove `colorable_fields`, this might get auto-inferred as a categorical field which would mean this test would no longer be testing what it's supposed to.

bmschmidt (changes requested):
Looks good, thanks. The only critical change here is the missing `memory_map` argument.

nomic/data_operations.py
rguo123: Makes sense. Is there any benefit to this reading in the bytes, reading it in with `feather.read_table` than writing it?

nomic/data_operations.py
```diff
@@ -846,3 +846,125 @@ def remove(self, ids: List[str], tags: List[str], delete_all: bool = False) -> b
...
+        tbs = []
+        root = feather.read_table(self.projection.tile_destination / "0/0/0.feather")
+        try:
+            small_sidecars = set(
+                [v for k, v in json.loads(root.schema.metadata[b"sidecars"]).items()]
+            )
+        except KeyError:
+            small_sidecars = []
+        for path in self.projection._tiles_in_order():
+            tb = pa.feather.read_table(path).select(["_id"])
```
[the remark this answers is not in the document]
rguo123: Are colorable fields not getting stored in sidecars? When we do `fetch_tiles`, I saw logic for writing sidecars as their own feather file so figured the `small_sidecars` bit would be getting this.

nomic/data_operations.py
```diff
@@ -846,3 +846,125 @@ def remove(self, ids: List[str], tags: List[str], delete_all: bool = False) -> b
...
+            for sidecar_file in small_sidecars:
+                carfile = pa.feather.read_table(
```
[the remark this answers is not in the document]
rguo123: done

nomic/data_operations.py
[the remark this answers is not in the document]
bmschmidt: Nope

AndriyMulyar:
@rguo123 unit tests are failing for a real reason here.

bmschmidt (approved):

PR #208, opened 2023-09-01
[RFR] Allow iterables, dataframes, list of dictionaries in map_text

rguo123:


apage43 (approved):
looks good, and I was able to use this to upload a text project straight from a HF `datasets` IterableDataset https://huggingface.co/docs/datasets/stream 

```python
dataset = load_dataset('oscar', "unshuffled_deduplicated_en", split='train', streaming=True)
atlas.map_text(name='oscarsubset', data=itertools.islice(dataset, 25_000), indexed_field='text', reset_project_if_exists=True)
```

nomic/atlas.py
```diff
@@ -165,12 +166,13 @@ def map_text(
     projection_spread: float = DEFAULT_PROJECTION_SPREAD,
     duplicate_detection: bool = False,
     duplicate_threshold: float = DEFAULT_DUPLICATE_THRESHOLD,
+    upload_batch_size: int = 1000
```
bmschmidt, line 169: This should not be a user-facing parameter. To choose a size anything other than haphazardly, a user would need to know details of our backend ingest system. We shouldn't make users choose parameters they can't set.

nomic/tests/test_atlas_client.py
```diff
@@ -382,3 +382,56 @@ def test_map_embeddings():
...
+        name='UNITTEST_pandas_text',
+        id_field='id',
+        indexed_field="color",
+        data=data_iter,
+        is_public=True,
+        colorable_fields=['color'],
+        reset_project_if_exists=True,
+    )
+
+    map = project.get_map(name='UNITTEST_pandas_text')
```
bmschmidt, line 435: Should not re-use the project name here for both tests with `reset_project_if_exists`; seems like pytest doesn't run in parallel by default, but maybe we'll use one of the packages to speed up the tests some day.

PR #213, opened 2023-09-11
Revert map_text without iterables but still make pandas dfs work

rguo123:


rguo123:
Closing this. Although pandas iterators are still messy with the casting, let's address when it breaks and keep pushing for people to use add_text direclty in cases of large data.

PR #214, opened 2023-09-20
feat: allow arrow table for upload in map_text

zanussbaum:
Previously, this failed as iterating over a pyarrow table iterates over the columns. For example, 

```
import pyarrow as pa

>>> schema = pa.schema([
...     ('text', pa.string()),
...     ('url', pa.string())
... ])
>>> 
>>> # Sample data
>>> data = {
...     'text': ['sample1', 'sample2', 'sample3', 'sample4', 'sample5'],
...     'url': ['http://example1.com', 'http://example2.com', 'http://example3.com', 'http://example4.com', 'http://example5.com']
... }
>>> 
>>> tb = pa.Table.from_pydict(data)
>>> tb
pyarrow.Table
text: string
url: string
----
text: [["sample1","sample2","sample3","sample4","sample5"]]
url: [["http://example1.com","http://example2.com","http://example3.com","http://example4.com","http://example5.com"]]
>>> for x in tb:
...     break
... 
>>> x
<pyarrow.lib.ChunkedArray object at 0x7fc8c663e7c0>
[
  [
    "sample1",
    "sample2",
    "sample3",
    "sample4",
    "sample5"
  ]
]
>>> 
```

when we want to iterate over the rows instead. Feel free to critique my pyarrow code, I just hacked around on this.

### DIFF
--- a/nomic/atlas.py
+++ b/nomic/atlas.py
@@ -6,6 +6,7 @@ or in a Jupyter Notebook to organize and interact with your unstructured data.
 import uuid
 from typing import Dict, Iterable, List, Optional, Union
 
+import pyarrow as pa
 import numpy as np
 from pandas import DataFrame
 from loguru import logger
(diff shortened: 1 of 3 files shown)


bmschmidt (approved):
LGTM. We may want to add a note at some point that it would faster and safer to insert arrow record batches directly rather than casting to list.

PR #221, opened 2023-10-03
[RFR] Topic Refactor 

rguo123:
Refactoring topic code to be compatible with new atlas cloud logic.

Implementation is backwards compatible so old maps will load topics correctly. Main change is topic fields contain topic ids that need to be joined with depth to be identifiable. Before this change, we were using the topic labels. Depths are also not guaranteed to be strictly (1, 2, 3).

`get_topic_density` also happens locally now. This requires downloading the time_field project field. Rather than downloading all the data using `AtlasMapData`, we include an optional `fields` parameter to just fetch the desired field.

TODO: one final test with staging lock for new topic code

### DIFF
--- a/nomic/data_operations.py
+++ b/nomic/data_operations.py
@@ -122,9 +122,6 @@ class AtlasMapTopics:
     topic_depth columns contain the topic id at that depth.
     Topic ids are unique for each depth.
     Corresponding labels can be found in `topics.metadata`.
-    
-    Backwards Compatibility Note:  
-    Before October 2023, topic_depth columns contained the topic label.
 
     === "Accessing Topics Example"
         ``` py
--- a/nomic/data_operations.py
+++ b/nomic/data_operations.py
@@ -119,9 +119,6 @@ class AtlasMapDuplicates:
 class AtlasMapTopics:
     """
     Atlas Topics State
-    topic_depth columns contain the topic id at that depth.
-    Topic ids are unique for each depth.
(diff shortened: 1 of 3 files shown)


nomic/data_operations.py
```diff
@@ -237,38 +246,39 @@ def group_by_topic(self, topic_depth: int = 1) -> List[Dict]:
                 topic_long_description, and list of datum_ids.
         """
 
-        topic_cols = []
         # TODO: This will need to be changed once topic depths becomes dynamic and not hard-coded
```
AndriyMulyar, line 249: remove this comment

AndriyMulyar (approved):

AndriyMulyar:
Approved with 1 comment and a question. Do we have to update the documentation examples found here: https://docs.nomic.ai/map_state/topics.html

The outputs have columns which are topic_depth_1, 2, etc. Now that depths are dynamic, how do we maintain such a mapping?

nomic/data_operations.py
rguo123: removed

rguo123:
I made some small documentation updates but no major changes as everything still functions similarly.
For dynamic topic depths, topic columns are either found in primary quad tiles (old version) or in their own sidecar. Both are loaded in `_fetch_tiles` which means we can search for columns that start with `_topic_depth` (reserved column name) to fetch them dynamically

AndriyMulyar (approved):

PR #222, opened 2023-10-09
Embedding inference (beta)

AndriyMulyar:


nomic/embed.py
```diff
@@ -0,0 +1,107 @@
...
+    response = requests.post(
+        atlas_class.atlas_api_path + "/v1/embedding/text",
+        headers=atlas_class.header,
+        json={'texts': texts, 'model': model},
+    )
+
+    if response.status_code == 200:
+        return response.json()
+    else:
+        raise Exception(str(response.json()))
```
bstadt, line 37: this throws an unintuitive error when the user is not logged in

Issue #235, opened 2023-11-11
List projects under my account


YasmineMh:
Hey, is there a method to list all projects under my Nomic account?
Thanks!

bmschmidt:
Unless I'm completely failing to find it, there is not--but there should be! We'll try to get something in the next release--I'll start off submitting the following function, which you can use in the meantime if it helps.

```py
from nomic.project import AtlasClass
import requests

def list_projects(organization_id=None):
    """
    Lists all projects in an organization.

    If called without an organization id, it will list all projects in the
    current user's main organization.
    """
    c = AtlasClass()
    if organization_id is None:
        organization = c._get_current_users_main_organization()
        if organization is None:
            raise ValueError(
                "No organization id provided and no main organization found."
            )
        organization_id = organization['organization_id']
    response = requests.get(
        c.atlas_api_path + f"/v1/organization/{organization_id}",
        headers=c.header
    )
    proj_info = response.json()['projects']
    return [
        {'name': p['project_name'],
            'id': p['id'],
            'created_timestamp': p['created_timestamp']}
        for p in proj_info
    ]

```

YasmineMh:
Thank you @bmschmidt ! it's working.

PR #236, opened 2023-11-11
allow listing projects for an organization

bmschmidt:
Addresses #235 .

I'm not sure if the test here makes sense--it ensures the method behaves correctly for *my* user, but in some cases the test user might not have any projects.

Does not include version bump etc., which I'll add after review.

### DIFF
--- a/nomic/tests/test_atlas_client.py
+++ b/nomic/tests/test_atlas_client.py
@@ -468,3 +468,10 @@ def test_map_text_iterator():
     map = project.get_map(name='UNITTEST_pandas_text')
     assert project.total_datums == 50
     project.delete()
+
+
+def test_list_projects():
+    projects = atlas.list_projects()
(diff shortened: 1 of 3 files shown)


nomic/tests/test_atlas_client.py
```diff
@@ -468,3 +468,10 @@ def test_map_text_iterator():
     map = project.get_map(name='UNITTEST_pandas_text')
     assert project.total_datums == 50
     project.delete()
+
+
+def test_list_projects():
+    projects = atlas.list_projects()
```
apage43, line 474: several of the other tests create projects, so I figure this one should also temporarily create a project if its testing something that will only work if at least one exists - it doesn't need to bother with creating an _index_

bmschmidt: Addressed

PR #240, opened 2023-11-21
basic custom endpoint support

apage43:


apage43:
dropped env vars in favor of a cli arg, infer that enterprise api domains are `api.` of the www domain, and changed to save domains to the credentials file after enteprise login

AndriyMulyar (approved):

PR #241, opened 2023-11-29
reuse previous credentials on bearer refresh

bmschmidt:
For enterprise edition because the command line flag is no longer around.


apage43 (approved):

apage43 (approved):

apage43:
my approval doesn't seem to allow merging here

PR #246, opened 2024-01-09
[RFR] Download tags

rguo123:
This PR updates AtlasMapTags to match the new atlas_cloud implementation.

Code changes include:
- Fetching completed tags
- Downloading bitmasks for tags
- Returning datum_ids associated with a tag
- Returning a single dataframe with all tags for each datum_id
- Deprecation: removes adding or removing tags using Python client

note:
I started renaming mentions of self.project to self.dataset for consistency. May not be appropriate place to do it and will remove if so

### DIFF
--- a/nomic/data_operations.py
+++ b/nomic/data_operations.py
@@ -751,7 +751,7 @@ class AtlasMapTags:
         #     tags: A list containing the tags you want to apply to these data points.
 
         # '''
-        raise NotImplementedError("AtlasMapTags.add is not supported.")
+        raise NotImplementedError("AtlasMapTags.add is currently not supported.")
 
     def remove(self, ids: List[str], tags: List[str], delete_all: bool = False) -> bool:
         # '''
@@ -766,7 +766,7 @@ class AtlasMapTags:
         #     True on success.
 
         # '''
-        raise NotImplementedError("AtlasMapTags.remove is not supported.")
+        raise NotImplementedError("AtlasMapTags.remove is currently not supported.")
 
     def __repr__(self) -> str:
         return str(self.df)
--- a/nomic/data_operations.py
+++ b/nomic/data_operations.py
@@ -596,7 +596,7 @@ class AtlasMapTags:
         tbs = []
         all_quads = list(self.projection._tiles_in_order(coords_only=True))
         for quad in tqdm(all_quads):
-            quad_str = os.path.join([str(q) for q in quad])
+            quad_str = os.path.join(*[str(q) for q in quad])
             datum_id_filename = quad_str + "." + "datum_id" + ".feather"
             path = self.projection.tile_destination / Path(datum_id_filename)
             tb = feather.read_table(path, memory_map=True)
@@ -694,7 +694,7 @@ class AtlasMapTags:
         all_quads = list(self.projection._tiles_in_order(coords_only=True))
         ordered_tag_paths = []
         for quad in tqdm(all_quads):
-            quad_str = os.path.join([str(q) for q in quad])
+            quad_str = os.path.join(*[str(q) for q in quad])
             filename = quad_str + "." + f"_tag.{tag_definition_id}" + ".feather"
             path = self.projection.tile_destination / Path(filename)
             download_attempt = 0
@@ -725,7 +725,7 @@ class AtlasMapTags:
         # NOTE: This currently only gets triggered on `df` property
         all_quads = list(self.projection._tiles_in_order(coords_only=True))
         for quad in tqdm(all_quads):
-            quad_str = os.path.join([str(q) for q in quad])
+            quad_str = os.path.join(*[str(q) for q in quad])
             tile = self.projection.tile_destination / Path(quad_str)
             tile_dir = tile.parent
             if tile_dir.exists():
@@ -840,7 +840,7 @@ class AtlasMapData:
 
         for quad in tqdm(all_quads):
             for sidecar in sidecars:
-                quad_str = os.path.join([str(q) for q in quad])
+                quad_str = os.path.join(*[str(q) for q in quad])
                 encoded_colname = base64.urlsafe_b64encode(sidecar.encode("utf-8")).decode("utf-8")
                 filename = quad_str + "." + encoded_colname + ".feather"
                 path = self.projection.tile_destination / Path(filename)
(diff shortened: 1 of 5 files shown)


nomic/utils.py
```diff
@@ -233,3 +236,12 @@ def get_object_size_in_bytes(obj):
...
     return sz
+
+# Helpful function for downloading feather files
+def download_feather(url: str, path: str):
+    path.parent.mkdir(parents=True, exist_ok=True)
+    with requests.get(url, stream=True) as response:
+        response.raise_for_status()
+        with open(path, 'wb') as f:
+            for chunk in response.iter_content(chunk_size=8192):
+                f.write(chunk)
```
bmschmidt, line 247: As a rule we *don't* want to use this function, so I think shoudl remove it. We store feather files on our web site without column level compression, but with gzip compression applied to the whole file, because the javascript arrow package does not support column level  compression. It's better to read the full arrow file into memory and then write it to disk for two reasons:
1. We can apply good zstd compression client side, using less disk space of our users;
2. It ensures that the files are valid arrow on disk, rather than let them potentially get written with a 'feather' suffix but actually be the text of a 403 error or something.

It's better to use the model in the quadtrees, because this will use zstd compression:

```python
                data = requests.get(url)
                readable = io.BytesIO(data.content)
                readable.seek(0)
                tb = feather.read_table(readable, memory_map=True)
                path.parent.mkdir(parents=True, exist_ok=True)
                feather.write_feather(tb, path)
```

BUT the one place I deliberately used the old `download_file` method was for getting sidecars, because it's possible that someone will have specifically a full text sidecar for say Obelics where even a single quadtree tile is too large to load into memory. For tags this isn't the case--we know they're small enough to load.

bmschmidt: A second issue here, because that one is getting long--these functions also need to include bearer tokens, don't they? Just a plain requests get won't be authenticated. I'm a little concerned in fact that the tests are *passing*--does that mean we're overly permissive in letting non-authenticated users download tags.

nomic/dataset.py
```diff
@@ -696,9 +696,9 @@ def _get_atoms(self, ids: List[str]) -> List[Dict]:
             raise ValueError("You must specify a list of ids when getting data.")
 
         response = requests.post(
-            self.project.atlas_api_path + "/v1/project/atoms/get",
-            headers=self.project.header,
-            json={'project_id': self.project.id, 'index_id': self.atlas_index_id, 'atom_ids': ids},
+            self.dataset.atlas_api_path + "/v1/project/atoms/get",
```
bmschmidt, line 699: I had forgot we had an endpoint literally called `get` that used `POST` instead of `GET` to retrieve data.

nomic/utils.py
rguo123: Ah agreed that we need to add bearer tokens. These methods currently dont have tests as I need to create a tagged map on the test account (see todo) so no need to be concerned about passing tests.

And gotcha that makes sense on reading and writing feather.

nomic/dataset.py
```diff
@@ -651,13 +643,21 @@ def _download_feather(self, dest: Optional[Union[str, Path]] = None, overwrite:
...
+
+            if not path.exists() or overwrite or should_download:
+                data = requests.get(root + quad)
                 readable = io.BytesIO(data.content)
                 readable.seek(0)
                 tb = feather.read_table(readable, memory_map=True)
                 path.parent.mkdir(parents=True, exist_ok=True)
                 feather.write_feather(tb, path)
+
             schema = ipc.open_file(path).schema
```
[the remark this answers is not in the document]
rguo123: yeah this should work. much cleaner, thanks!

nomic/utils.py
rguo123: Adding on, this method didn't exist separately in utils until this PR. I noticed that I was using it in a few places so I pulled it out. I don't think we should remove it but keep a general method that uses the feather read and write

nomic/data_operations.py
```diff
@@ -641,130 +574,194 @@ class AtlasMapTags:
...
+        # Args:
+        #     ids: The datum_ids to delete tags from.
+        #     tags: The list of tags to delete from the data points. Each tag will be applied to all data points in `ids`.
+        #     delete_all: If true, ignores ids parameter and deletes all specified tags from all data points.
+
+        # Returns:
+        #     True on success.
+
+        # '''
+        raise NotImplementedError("AtlasMapTags.remove is not supported.")
```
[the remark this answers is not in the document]
rguo123: I thought the reason we are going for NotImplemented is because we may/probably will bring it back at some point. Dont mind changing it to deprecated but I think NotImplemented is better as this only impacts users that have been using Python client to add/remove tags and if they see a DeprecatedWarning and then it comes back later it's _slightly_ more confusing than NotImplemented.

nomic/dataset.py
```diff
@@ -640,25 +632,39 @@ def _download_feather(self, dest: Optional[Union[str, Path]] = None, overwrite:
...
+        quads = deque([f'0/0/0'])
         self.tile_destination.mkdir(parents=True, exist_ok=True)
-        root = f'{self.project.atlas_api_path}/v1/project/{self.project.id}/index/projection/{self.id}/quadtree/'
-        quads = [f'0/0/0']
+        root = f'{self.dataset.atlas_api_path}/v1/project/{self.dataset.id}/index/projection/{self.id}/quadtree/'
         all_quads = []
         sidecars = None
         while len(quads) > 0:
-            rawquad = quads.pop(0)
+            rawquad = quads.popleft()
```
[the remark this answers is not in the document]
rguo123: Ah yeah, let me remove. I started with deque first and realized the retries was a better way of doing it

nomic/data_operations.py
```diff
@@ -641,130 +574,199 @@ class AtlasMapTags:
...
+        tags = self.get_tags()
+        tag_definition_ids = [tag["tag_definition_id"] for tag in tags]
+        if self.auto_cleanup:
+            self._remove_outdated_tag_files(tag_definition_ids)
+        for tag in tags:
+            self._download_tag(tag["tag_name"], overwrite=overwrite)
+        tbs = []
+        all_quads = list(self.projection._tiles_in_order(coords_only=True))
+        for quad in tqdm(all_quads):
+            quad_str = "/".join([str(q) for q in quad])
```
bmschmidt, line 599: Use pathlib instead of `"/".join` to create safe filesystem paths.

rguo123: yeppppp same mistake that i made last time 🤦. fixed

bmschmidt (approved):

PR #254, opened 2024-01-11
v2.x hotfix for enteprise auth without cli

apage43:
note this is not a PR to main, but to a `v2.x` branch I just created from the 2.0.16 "version bump" commit

- fix login() for enterprise
- version bump (2.0.17)


apage43:
merged to v2.x since its now deployed

PR #276, opened 2024-03-18
Overwrite to False for downloading tiles?

rguo123:


bmschmidt:
@apage43 said that we might be able to use modified headers for this--I looked at the headers we send back from fastAPi and couldn't see modified time. But would be preferable to check if possible, maybe he can weigh in?

apage43:
> @apage43 said that we might be able to use modified headers for this--I looked at the headers we send back from fastAPi and couldn't see modified time. But would be preferable to check if possible, maybe he can weigh in?

Anything we return as a FileResponse (such as the quadtiles) _should_ be sent with a `Last-Modified` header, as well as an `ETag`:

```
curl -I https://api-atlas.nomic.ai/v1/project/77f236d4-8230-4427-ab72-93e1a0c760ea/index/projection/71ee5201-27f7-434b-b8db-62bcc3ecc253/quadtree/0
/0/0.feather
HTTP/2 200
content-type: application/octet-stream
content-length: 84514
date: Sat, 23 Mar 2024 21:08:58 GMT
server: nginx/1.23.3
last-modified: Thu, 14 Mar 2024 16:49:51 GMT
etag: ee6908d72dd172992f0d89c37e702991
```

ETags are meant to be treated as an opaque cache key (though using the last-modified header as one would also work fine) - if you remember the ETag header you got when you downloaded a file you can send a HEAD request to check if the remote ETag is still the same, if it is then you know the file has not changed. 

Not importat to know this, but its usually an md5 hash but not necessarily an md5 hash of the file _content_ - in Starlette/FastAPI its actually just `md5(f'{last_modified_time}-{file_size}')`

bmschmidt (approved):
Approving--we should really do this. There are some other ways to massively increase the speed of the client we should look into.

rguo123:
Ok I'm going to merge for now but create a follow-up ticket that we can do this more smartly with Last-Modified

PR #278, opened 2024-03-19
[RFR] Sagemaker Utils

rguo123:
This PR includes methods to 
- call sagemaker for real time inference and batch transform

We are including these methods to make it easy for users to call sagemaker endpoint as suggested by AWS onboarding doc.


nomic/aws/sagemaker.py
```diff
@@ -0,0 +1,181 @@
...
+    # padding and truncation are handled by tokenizer
+
+    all_tokens = tokenizer.encode(text, add_special_tokens=add_special_tokens).ids
+    if len(all_tokens) == 0:
+        logger.warning(f"Zero tokens generated from text.")
+        all_tokens = tokenize_text(EMPTY_PLACEHOLDER, add_special_tokens=False).ids
+    return all_tokens
+
+
+def create_sm_request_for_batch(texts: List[str], tokenizer: Tokenizer):
```
zanussbaum, line 127: nit: it took me a second to figure out what `sm` stands for, could rename to something like `create_sagemaker_request`

nomic/aws/sagemaker.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+def load_tokenizer(model: NomicTextEmbeddingModel) -> Tokenizer:
+    """
+    Instantiate the huggingface tokenizer.
+    :param model: the model doing the embedding
+    :return:
+    """
+    tokdir = Path(__file__).parent / "tokenizers"
+    if model == NomicTextEmbeddingModel.nomic_embed_text_v1_5:
```
zanussbaum, line 76: are we only supporting 1.5? what about v1?

nomic/aws/sagemaker.py
```diff
@@ -0,0 +1,181 @@
...
+    Args:
+        texts: List of texts to be batched.
+        tokenizer: Tokenizer to use.
+
+    Returns:
+        HTTP Request object for Triton server.
+    """
+    all_tokens = []
+    for text in texts:
+        all_tokens.append(tokenize_text(text, tokenizer=tokenizer))
```
zanussbaum, line 140: why are we tokenizing text like this? Why not just do 
```python
encoded = tokenizer.encode_batch(texts)
```
https://huggingface.co/docs/tokenizers/v0.13.4.rc2/en/quicktour#encoding-multiple-sentences-in-a-batch

nomic/aws/sagemaker.py
```diff
@@ -0,0 +1,181 @@
...
+    input_ids = np.array(all_tokens, dtype=np.int32)
+    mlen = max(len(tokens) for tokens in all_tokens)
+    attention_mask = [
```
zanussbaum, line 143: I think this also returns the attention mask

nomic/aws/__init__.py
```diff
@@ -0,0 +1,11 @@
+def _check_aws_dependencies():
+    try:
+        import tokenizers
+        import tritonclient
+    except ImportError as e:
+        missing_package = str(e).split("No module named ")[-1].strip("'")
+        raise ImportError(f"The '{missing_package}' package is required for this feature. "
+                          "Please install it by running 'pip install nomic[aws]'.")
+    
+_check_aws_dependencies()
```
zanussbaum, line 10: is this always run?

rguo123: only on importing this submodule

rguo123: import nomic will work fine
import nomic.aws will raise the ImportError

nomic/aws/sagemaker.py
rguo123: Just v1.5 for now. We will need to upload a new sagemaker endpoint for each version. As i understand it, we'll stick to one and get it working for marketplace/mongo before doing more

nomic/aws/sagemaker.py
rguo123: I almost want to make it hidden so people just use the generator but agreed on rename. thanks!

nomic/aws/sagemaker.py
apage43: this was avoided in the other code to in order to implement the chunk-and-mean behavior such that chunking happened at the token level - hence disabling padding/specials to encode _just_ the user input tokens in order to chunk them, then manually adding the specials and framing back to each chunk, so that _all_ chunks would get the correct framing and prefix (that is: `[CLS]search_document:{chunk_tokens}[SEP]`), which is why the mask was also manual

if we're just providing the model "raw" here and not trying to automatically do handling of prefixing/chunking etc. encode_batch is fine

nomic/aws/sagemaker.py
```diff
@@ -0,0 +1,197 @@
...
+    ):
+        response = client.invoke_endpoint(
+            EndpointName=sagemaker_endpoint,
+            Body=embed_request,
+            ContentType="application/vnd.sagemaker-triton.binary+json;json-header-size={}".format(
+                header_length
+            ),
+        )
+        embeddings.append(parse_sagemaker_response(response))
+    return np.vstack(embeddings)
```
zanussbaum, line 197: it might be good to document somewhere that this returns `np.float16` dtyp'd arrays

zanussbaum (approved):

nomic/aws/sagemaker.py
```diff
@@ -0,0 +1,197 @@
...
+
+    # have to set to binary since http doesn't natively support fp16
+    outputs.append(aiohttpclient.InferRequestedOutput("embedding", binary_data=True))
+    request, header_length = InferenceServerClient.generate_request_body(
+        inputs=inputs, outputs=outputs
+    )
+    return request, header_length
+
+
+def batch_sagemaker_requests(texts: List[str], batch_size=32):
```
apage43, line 122: make sure we test the longest text length we'll allow with the batch size we specify here

apage43 (approved):

apage43 (changes requested):
ah, you're importing the asyncio triton client but trying to use it outside of async fns, that won't work

nomic/aws/sagemaker.py
```diff
@@ -0,0 +1,197 @@
...
+
+import hashlib
+import logging
+from enum import Enum
+from pathlib import Path
+from typing import List, Optional
+
+import boto3
+import numpy as np
+import tritonclient.http.aio as aiohttpclient
```
apage43, line 13: you want the non-asyncio version (just drop the `.aio` and the api is the same, sans-`await`)

nomic/aws/sagemaker.py
rguo123: i thought we truncated the tokens?

nomic/aws/sagemaker.py
rguo123: added in docstring but i'll also specify in the tutorial notebook

nomic/aws/sagemaker.py
```diff
@@ -0,0 +1,161 @@
...
+import numpy as np
+import sagemaker
+from tqdm import tqdm
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+# NOTE: Currently Sagemaker only supports nomic-embed-text-v1.5 model.
+
+SAGEMAKER_MODELS = {"nomic-embed-text-v1.5": {"us-east-2": "TODO: ARN"}}
```
etaldot, line 15: should this be hardcoded. More of a question

nomic/aws/sagemaker.py
```diff
@@ -0,0 +1,161 @@
...
+        wait=wait,
+        logs=logs,
+    )
+    job_name = None
+    if embedder.latest_transform_job is not None:
+        job_name = embedder.latest_transform_job.name
+    return job_name
+
+
+def embed_texts(
```
zanussbaum, line 133: Note: we need to update this to work with prefixes

zanussbaum (approved):
LGTM, other than we need to add `prefix` as a parameter to the `embed_text` function

nomic/aws/sagemaker.py
```diff
@@ -0,0 +1,161 @@
...
+        role: Arn of the IAM role to use (must have permissions to S3 data as well).
+        max_payload: The maximum payload size in megabytes.
+        instance_type: The instance type to use.
+        n_instances: The number of instances to use.
+        wait: Whether method should wait for job to finish.
+        logs: Whether to log the job status (only meaningful if wait is True).
+    Returns:
+        The job name.
+    """
+    if arn is None:
```
etaldot, line 84: Nit but can we move these checks in a helper like validate_batch_transform_params

nomic/dataset.py
```diff
@@ -671,7 +700,7 @@ def _download_large_feather(self, dest: Optional[Union[str, Path]] = None, overw
             elif sidecars is None:
                 sidecars = set()
             if not "." in rawquad:
-                for sidecar in sidecars:
+                for sidecar in sidecars | registered_sidecars:
```
etaldot, line 703: is | and `or` one and the same. Also a question

etaldot (changes requested):
left few minor nits. Happy to make them as well if reasonable

AndriyMulyar:
Please post comments here. Merging so we have a static notebook link to share.

 https://linear.app/nomic/issue/NOM-1213/nomic-embed-sagemaker-updates
@etaldot @zanussbaum

nomic/dataset.py
AndriyMulyar: unrelated, these changes were rebased in not sure why github shows the diff.

apage43: > is | and `or` one and the same. Also a question

even though its unrelated, they are not. `or` is equivalent to `||`, not `|`.  `|` is either bitwise-or (for numbers) or in this case, union of sets

zanussbaum:
Only comments are:
* Update ARN 
* Allow prefixing for `embed_text`

PR #284, opened 2024-04-02
Expose created timestamp

AndriyMulyar:
Exposes created timestamp on an AtlasDataset.

### DIFF
--- a/nomic/dataset.py
+++ b/nomic/dataset.py
@@ -949,6 +949,10 @@ class AtlasDataset(AtlasClass):
         return self.meta['unique_id_field']
 
     @property
+    def created_timestamp(self) -> datetime:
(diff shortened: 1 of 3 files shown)


nomic/dataset.py
```diff
@@ -948,6 +948,10 @@ def id(self) -> str:
     def id_field(self) -> str:
         return self.meta['unique_id_field']
 
+    @property
+    def created_timestamp(self) -> datetime:
+        return datetime.strptime(self.meta['created_timestamp'], '%Y-%m-%dT%H:%M:%S.%f%z')
```
bmschmidt, line 953: This is iso 8601 right?

bmschmidt (approved):

PR #285, opened 2024-04-04
refactor: default model as None

wilsonjr:


nomic/data_inference.py
```diff
@@ -77,7 +76,7 @@ class NomicProjectOptions(BaseModel):
     n_epochs: Optional[int] = None
     spread: Optional[float] = None
     local_neighborhood_size: Optional[int] = None
-    model: str = DEFAULT_PROJECTION_MODEL
```
etaldot, line 80: assuming we do this since pydantic in api already provides a default correct?

etaldot (approved):

nomic/data_inference.py
wilsonjr: Yep that's it! Last deployment covers this

PR #286, opened 2024-04-05
embedding updates

zanussbaum:


nomic/embed.py
```diff
@@ -104,6 +104,34 @@ def text(texts: List[str], model: str = "nomic-embed-text-v1", task_type: str =
     return combined
 
 
+def image_api_request(images: List[Tuple[str, bytes]], model: str = 'nomic-embed-vision-v1'):
```
bmschmidt, line 107: Rather than do global scoping, could just instantiate the default atlas class in the function definition? All invocations of the function will use the same class because of python's kinda weird default arg behavior, but also it will be possible to override if we're in a case where we need to later.

```suggestion
def image_api_request(images: List[Tuple[str, bytes]], model: str = 'nomic-embed-vision-v1', atlas_class = AtlasClass()):
```

zanussbaum: I kept it using the global to keep it similar to the text implementation but I don't have any strong opinions. Maybe @apage43 has some thoughts?

apage43: the main reason that the global is initialized to None and only instantiated lazily 

(this section)
```python
    if atlas_class is None:
        atlas_class = AtlasClass()
```

is so that you have the opportunity to call `cli.login(api_key)` to log in _before_ you start trying to do embeddings so that you can use the client in notebook environments where you may not have the opportunity to log in with the CLI or otherwise don't want to store your API key on the local file system

Just putting `atlas_class=AtlasClass()` as a default arg instead would break that again, since default arg expressions are run when the `def` is evaluated (that is, _when you `import` the file_) - it needs to be deferred because instantiating  `AtlasClass()` will throw if you have not logged in yet (either via `cli.login` or the cli itself).

The reason not to just instantiate a *fresh* `AtlasClass()` every time this is called is because doing so does a call to the `/v1/user` endpoint to check that your credentials are correct - which we only should need to do once

nomic/embed.py
```diff
@@ -115,75 +143,46 @@ def images(images: Union[str, PIL.Image.Image], model: str = 'nomic-embed-vision
...
+    max_workers = 10
+    chunksize = MAX_IMAGE_REQUEST_SIZE
+    smallchunk = max(1, int(len(images) / max_workers))
+    # if there are fewer texts per worker than the max chunksize just split them evenly
+    chunksize = min(smallchunk, chunksize)
+
+    image_batch = []
+    for image in images:
+        if isinstance(image, str) and os.path.exists(image):
                 img = PIL.Image.open(image)
```
apage43, line 160: as-is we wouldn't resize images passed as file names - prolly want to wrap this in resize_pil to be consistent

nomic/embed.py
```diff
@@ -104,6 +104,34 @@ def text(texts: List[str], model: str = "nomic-embed-text-v1", task_type: str =
...
+    #if image is too large, downsample before sending over the wire
+    max_width = 512
+    max_height = 512
+    if max_width > 512 or max_height > 512:
+        downsize_factor = max(width/max_width, height/max_height)
+        img.resize((width/downsize_factor, height/downsize_factor))
+    return img
+
+
 def images(images: Union[str, PIL.Image.Image], model: str = 'nomic-embed-vision-v1'):
```
apage43, line 135:
```suggestion
def images(images: List[Union[str, PIL.Image.Image]], model: str = 'nomic-embed-vision-v1'):
```

apage43 (approved):

nomic/embed.py
apage43: if you really wanted to keep things out of the global namespace and be able to override the passed AtlasClass (though I don't know when we'd need to?), 

you could do something like this: https://gist.github.com/apage43/e8da6a7c6613b1cd2242ef761cddf7ae to encapsulate the lazy initialization logic and then you could use a LazyCallOnce as a default arg

AndriyMulyar (approved):

PR #287, opened 2024-04-06
embed.text: implement infererence_mode="local" based on Embed4All

cebtenzzre:


nomic/embed.py
```diff
@@ -61,18 +76,104 @@ def text_api_request(texts: List[str], model: str, task_type: str, dimensionalit
         raise Exception((response.status_code, response.text))
 
 
-def text(texts: List[str], model: str = "nomic-embed-text-v1", task_type: str = "search_document", dimensionality: int = None, long_text_mode: str = "truncate"):
+@overload
+def text(
+    texts: list[str],
```
zanussbaum, line 81: nit: should we use `List` instead of `list`? IIRC python < 3.9 doesn't support typing using the `list` keyword

nomic/embed.py
```diff
@@ -61,18 +76,104 @@ def text_api_request(texts: List[str], model: str, task_type: str, dimensionalit
...
-        texts: the text to embed
-        model: the model to use when embedding
-        task_type: the task type to use when embedding. One of `search_query`, `search_document`, `classification`, `clustering`
+        texts: The text to embed.
+        model: The model to use when embedding.
+        task_type: The task type to use when embedding. One of `search_query`, `search_document`, `classification`, `clustering`.
+        dimensionality: The embedding dimension, for use with Matryoshka-capable models. Defaults to full-size.
+        long_text_mode: How to handle texts longer than the model can accept. One of `mean` or `truncate`.
+        inference_mode: How to generate embeddings. One of `remote` or `local` (Embed4All). Defaults to `remote`.
+        device: The device to use for local embeddings. Defaults to CPU, or Metal on Apple Silicon. It can be set to:
```
zanussbaum, line 134: should we log or print something that we ignore `device` when `inference_mode` is `atlas`?

zanussbaum (commented):
some small comments

nomic/embed.py
cebtenzzre: I added `from __future__ import annotations` so these annotations are just strings and don't have to be valid at runtime. This module imports fine with python 3.8.

zanussbaum: ah TIL thank you!

zanussbaum (approved):

bmschmidt (commented):
General questions:

1. What is the behavior here if I pass in a single string of length 10,000 tokens?
2. What guarantees of consistency do we want to have between the cloud service and nomic embed local? If I put that long string in to both places, will I get the same value back? What about at 4096 tokens (our context window is 8192, but IIRC we chop up into smaller chunks than that in atlas_cloud--cc @apage43 )

bmschmidt:
Third question is whether this should include tests.

In general though this looks a good valuable addition, thanks @cebtenzzre !

cebtenzzre:
> 1. What is the behavior here if I pass in a single string of length 10,000 tokens?

With default parameters (long_text_mode=truncate), the input will be truncated to 2048 tokens, including the prefix.

> 2. What guarantees of consistency do we want to have between the cloud service and nomic embed local? If I put that long string in to both places, will I get the same value back?

The Embed4All backend has been updated to be fully compatible with the chunking and truncation behavior of the Nomic embedding API, especially when Embed4All is used with atlas=True. In the testing I've done, the result has always been the same within a very small margin of error (mostly due to floating point inaccuracies, with a few very minor tokenizer differences for certain non-latin characters that are arguably bugs on the HF side).

> What about at 4096 tokens (our context window is 8192, but IIRC we chop up into smaller chunks than that in atlas_cloud--cc @&ZeroWidthSpace;apage43 )

The input will still be truncated to 2048 tokens. The longest input you can use with *long_text_mode=mean* is 8192 tokens (excluding the prefix); any more than that and an error will be raised.

cebtenzzre:
I added some tests for embed.text, and fixed a few bugs. Outstanding issues:

1. Increasing n_ctx doesn't actually allow any more sequences to be embedded in one forward pass of Embed4All.

2. Attempting to initialize Embed4All on the GPU more than once crashes. This could previously be reproduced by calling embed.text with a non-CPU device and changing the Embed4All kwargs:
```python
from nomic import embed
embed.text(['x'], inference_mode='local', device='gpu')
embed.text(['x'], inference_mode='local', device='gpu', n_ctx=4096)
```
For now, we raise NotImplementedError in this case.

3. Embedding an empty string with Embed4All crashes. For now, we raise NotImplementedError in this case. Related to ggerganov/llama.cpp#6498, which also introduces a breaking change for the Nomic Embed GGUFs and has not yet been included in Embed4All.

bmschmidt (approved):

PR #288, opened 2024-04-10
fix: dataset sidecar download - datum id sidecars are special

apage43:
Downloads the id column when fetching map data, it got lost once sidecars stopped being registered on the root tile - the id column is _special_, it isn't considered part of the "dataset columns" and is always stored as a sidecar named "`datum_id`", _not_ base64'd like user supplied columns (with the original column name in the underlying arrow file) - we still want it here since it is a user supplied column

Fixes NOM-1294

### DIFF
--- a/nomic/data_operations.py
+++ b/nomic/data_operations.py
@@ -817,7 +817,11 @@ class AtlasMapData:
                 for col in carfile.column_names:
                     tb = tb.append_column(col, carfile[col])
             for big_sidecar in additional_sidecars:
-                fname = base64.urlsafe_b64encode(big_sidecar.encode("utf-8")).decode("utf-8")
+                fname = (
+                    base64.urlsafe_b64encode(big_sidecar.encode("utf-8")).decode("utf-8")
+                    if big_sidecar != 'datum_id'
+                    else big_sidecar
+                )
                 carfile = pa.feather.read_table(path.parent / f"{path.stem}.{fname}.feather", memory_map=True)
                 for col in carfile.column_names:
                     tb = tb.append_column(col, carfile[col])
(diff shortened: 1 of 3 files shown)


nomic/data_operations.py
```diff
@@ -844,18 +849,22 @@ def _download_data(self, fields=None):
         else:
             for field in sidecars:
                 assert field in self.dataset.dataset_fields, f"Field {field} not found in dataset fields."
+        encoded_sidecars = [base64.urlsafe_b64encode(sidecar.encode("utf-8")).decode("utf-8") for sidecar in sidecars]
+        if any(sidecar == 'datum_id' for (field, sidecar) in registered_sidecars):
+            sidecars.append('datum_id')
+            encoded_sidecars.append('datum_id')
+        print(f'{sidecars=}, {registered_sidecars=}')
```
bmschmidt, line 856: Is this intended, or is it debug code?

bmschmidt (approved):
Makes sense, with one question about a print statement. 

Noting though that eventually we should move to using the data columns endpoint, same as in the website, because it works off the list of what's been written rather than what theoretically should exist. Can't do that yet though because it has a misspelling for _embeddings in it.

bmschmidt:
Also noting that the only reason datum id is special is that it's the oldest sidecar, I hadn't realized we'd need base64 encoding when I made it. We could move to base64 encoding the colname for it eventually if we want.

nomic/data_operations.py
apage43: whoops, that was debug code

apage43, line 856:
```suggestion
```

AndriyMulyar (approved):

AndriyMulyar:
need a version bump here too

PR #292, opened 2024-04-12
Ensure code formatter

wilsonjr:


nomic/embed.py
```diff
@@ -132,17 +154,17 @@ def run_inference(batch):
...
-            downsize_factor = max(width/max_width, height/max_height)
-            img.resize((width/downsize_factor, height/downsize_factor))
+            downsize_factor = max(width / max_width, height / max_height)
+            img.resize((width / downsize_factor, height / downsize_factor))
         return img
 
     def send_request(i):
         image_batch = []
-        shard = images[i:i+IMAGE_EMBEDDING_BATCH_SIZE]
+        shard = images[i : i + IMAGE_EMBEDDING_BATCH_SIZE]
```
cebtenzzre, line 167: I'm not personally a fan of spaces around the `:` in a slice.

AndriyMulyar: yeah agreed, maybe disable this.

.circleci/config.yml
```diff
@@ -4,12 +4,19 @@ commands:
...
+        - run:
+            name: Update
+            command: sudo apt update
+        - run:
+            name: Install NPM
+            command: sudo apt install nodejs npm
         - run:
             name: Install python3 and pip3
-            command: sudo apt-get update && sudo apt-get -y install python3 python3-pip
+            command: sudo apt-get update && sudo apt-get -y install python3 python3-pip && sudo npm install
```
AndriyMulyar, line 15: why are we installing npm

cebtenzzre: On one hand, pyright is a node.js package. On the other hand, we're installing it via pip and I believe that it is independent of the system node.js/npm because it depends on [nodeenv](https://pypi.org/project/nodeenv/).

nomic/embed.py
cebtenzzre: black is doing this, and it's not configurable.

AndriyMulyar (commented):
Please unify the build path for CI and local dev in the [dev] set of requirements in circleci. Also, add a global linting command to the makefile with something like this

```
lint: black isort pyright
	@echo "Lint checks passed!"
```

AndriyMulyar (approved):

PR #295, opened 2024-04-18
Sagemaker Cleanup

rguo123:
-cleanup unneeded methods
-add prefixes
-match response from atlas embedding endpoint
-update example

### DIFF
--- a/nomic/aws/sagemaker.py
+++ b/nomic/aws/sagemaker.py
@@ -163,6 +163,6 @@ def embed_texts(
 
     return {
         "embeddings": embeddings,
-        "model": sagemaker_endpoint,
+        "model": "nomic-embed-text-v1.5",
         "usage": {},
     }
--- a/nomic/aws/sagemaker.py
+++ b/nomic/aws/sagemaker.py
@@ -139,6 +139,7 @@ def embed_texts(
         texts: List of texts to be embedded.
         sagemaker_endpoint: The sagemaker endpoint to use.
         region_name: AWS region sagemaker endpoint is in.
+        task_type: The task type to use when embedding.
         batch_size: Size of each batch. Default is 32.
 
     Returns:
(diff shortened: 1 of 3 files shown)


nomic/aws/sagemaker.py
```diff
@@ -140,15 +139,18 @@ def embed_texts(
         texts: List of texts to be embedded.
         sagemaker_endpoint: The sagemaker endpoint to use.
         region_name: AWS region sagemaker endpoint is in.
-        batch_size: Size of each batch.
```
zanussbaum, line 143: nit: should add `task_type` to docstring

nomic/aws/sagemaker.py
```diff
@@ -157,5 +159,10 @@ def embed_texts(
         response = client.invoke_endpoint(
             EndpointName=sagemaker_endpoint, Body=batch, ContentType="application/json"
         )
-        embeddings.append(parse_sagemaker_response(response))
-    return np.vstack(embeddings)
+        embeddings.extend(parse_sagemaker_response(response))
+
+    return {
+        "embeddings": embeddings,
+        "model": "nomic-embed-text-v1.5",
```
zanussbaum, line 166: If we are using v1.5, we can and should support both binary and matryoshka embeddings

rguo123: ok i can do this in nomic client but I think it would be better to make the change in sagemaker service.
i think merge this in as-is and i'll include those options in separate PR once i figure out how to push updates to marketplace offering.

nomic/aws/sagemaker.py
rguo123: done, thanks for catch!

zanussbaum (approved):
with condition that we fix matryoshka soon

nomic/aws/sagemaker.py
rguo123: https://linear.app/nomic/issue/NOM-1368/add-matryoshka-and-binary-option-to-sagemaker-service

PR #296, opened 2024-04-19
Matryoshka and Binary Options for Sagemaker

rguo123:
Adds in options to get binary or reduced dimensions from returned sagemaker embeddings.
Updates example notebook to show binary and matryoshka options.

Note: I did not add this in for batch transform as it would require the user to know how to format their csvs to pass the parameters.

### DIFF
--- a/nomic/aws/sagemaker.py
+++ b/nomic/aws/sagemaker.py
@@ -15,9 +15,7 @@ def _get_sagemaker_role():
     try:
         return sagemaker.get_execution_role()
     except ValueError:
-        raise ValueError(
-            "Unable to fetch sagemaker execution role. Please provide a role."
-        )
+        raise ValueError("Unable to fetch sagemaker execution role. Please provide a role.")
 
 
 def parse_sagemaker_response(response):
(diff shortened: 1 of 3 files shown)


nomic/aws/sagemaker.py
```diff
@@ -131,6 +131,8 @@ def embed_texts(
     region_name: str,
     task_type: str = "search_document",
     batch_size: int = 32,
+    dimensions: int = 768,
```
zanussbaum, line 134: to match the api, we should make this `dimensionality`

zanussbaum (commented):
How come we can't update the `batch_transform`? I tried to do a little digging on what the csv should be like but I couldn't find any definitive docs. 

Using the CSV, can we accept the dimensionality input? If so, can we provide an example to show how you can do batch transform? I think it would be useful

rguo123:
@zanussbaum we can add! There is no csv standard from what I can tell. We just need to be able to handle the expected format. It's already somewhat unique from old sagemaker given that a user has to know to upload a single column csv of texts in our case. 

What I can add is an option to have the first row be columns of parameter values. We'll check for this condition and if true, we'll attempt to parse the dimensionality and binary option. If it fails, we can catch the exception and treat is as a text row.

How does that sound to you?

zanussbaum:
yeah that sounds great! my only request is that we document this somewhere either in the doc strings or in an example readme somewhere so it's clear :D

zanussbaum (approved):
LGTM

PR #299, opened 2024-04-26
More black

bmschmidt:
This does a few things

1. Moves black and isort to `py39` instead of `py36` because we identified that as our target version
2. Creates a `pyproject.toml` with more opinionated linting standards around string normalization.
    ```toml
    [tool.black]
    line-length = 120
    target-version = ['py39']
    skip-string-normalization = false
    
    [tool.isort]
    profile = "black"
    skip = ["env", "ci_venv"]
    ignore_whitespace = false
    atomic = true
    line_length = 120
    ```
3. Changes the makefile to just use this linting config, rather than having cli flags manually specifying things like '120 lines' all over the place in places where they might get lost.
4. Actually applies these changes
5. Adds a git-blame-ignore-revs file to not blame me for all these changes, and edits the `make dev` script to ensure that local gitconfigs see that file.


AndriyMulyar:
Wilson should be the reviewer for this and run this locally to verify.

makefile
```diff
@@ -12,13 +12,13 @@ venv:
...
+	git config blame.ignoreRevsFile .git-blame-ignore-revs
 
 black:
-	source env/bin/activate; black -l 120 -S --target-version py36 nomic
-
+	source env/bin/activate; black nomic
 
 isort:
-	source env/bin/activate; isort  --ignore-whitespace --atomic -w 120 nomic
+	source env/bin/activate; isort --ignore-whitespace --atomic nomic
```
bmschmidt, line 21:
```suggestion
	source env/bin/activate; isort nomic
```

AndriyMulyar (approved):

PR #302, opened 2024-05-17
embed: fix regression in GPT4All detection

cebtenzzre:
Since #292 (released in 3.0.25, as 3.0.24 doesn't exist), the nomic client always assumes GPT4All is installed and fails to display `The 'gpt4all' package is required for local inference.`, instead erroring later when it fails to find `CancellationError`. This is because `elif embed4all_installed is None:` does not work, because embed4all_installed is always True or False.

This PR fixes that by restoring the original pattern, and adding an additional TYPE_CHECKING guard.

I also fixed a few recent additions of `if isinstance(...)` that made no sense - the pattern is supposed to be `assert isinstance(...)`, otherwise you are potentially introducing a bug. They were made unnecessary by adding a proper set of overloads for AtlasProjection._tiles_in_order (since you aren't supposed to return a `Union` if the return type can be assumed by the caller, and `Any` seemed too broad for this simple case).

One thing I don't understand is why #292 changed *some* of the PEP 604 unions to the old-style Optional, but left others alone. If PEP 604 is not allowed in this repo, shouldn't *all* of the new-style unions be removed? What about PEP 585 syntax, as also used in this file?

### DIFF
--- a/nomic/data_operations.py
+++ b/nomic/data_operations.py
@@ -434,19 +434,18 @@ class AtlasMapEmbeddings:
 
         for path in self.projection._tiles_in_order(coords_only=False):
             # double with-suffix to remove '.embeddings.feather'
-            if isinstance(path, Path):
-                files = path.parent.glob(path.with_suffix("").stem + "-*.embeddings.feather")
-                # Should there be more than 10, we need to sort by int values, not string values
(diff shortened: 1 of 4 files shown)


zanussbaum (approved):
LGTM

PR #304, opened 2024-05-31
[wip] Safe download of embeddings

rguo123:
Download embeddings using existing sidecar routes rather than a separate endpoint. 
Use download utils that has some guarantees about cleaning up bad files and doing retries.

### DIFF
--- a/nomic/data_operations.py
+++ b/nomic/data_operations.py
@@ -706,20 +706,7 @@ class AtlasMapTags:
             quad_str = os.path.join(*[str(q) for q in quad])
             filename = quad_str + "." + f"_tag.{tag_definition_id}" + ".feather"
             path = self.projection.tile_destination / Path(filename)
-            download_attempt = 0
-            download_success = False
-            while download_attempt < 3 and not download_success:
-                download_attempt += 1
-                if not path.exists() or overwrite:
-                    download_feather(root_url + filename, path, headers=self.dataset.header)
-                try:
-                    ipc.open_file(path).schema
-                    download_success = True
-                except pa.ArrowInvalid:
-                    path.unlink(missing_ok=True)
-
-            if not download_success:
-                raise Exception(f"Failed to download tag {tag_name}.")
+            download_feather(root_url + filename, path, headers=self.dataset.header, overwrite=True)
             ordered_tag_paths.append(path)
         return ordered_tag_paths
 
@@ -860,10 +847,8 @@ class AtlasMapData:
                 quad_str = os.path.join(*[str(q) for q in quad])
                 filename = quad_str + "." + encoded_colname + ".feather"
                 path = self.projection.tile_destination / Path(filename)
-
-                if not os.path.exists(path):
-                    # WARNING: Potentially large data request here
-                    download_feather(root + filename, path, headers=self.dataset.header)
+                # WARNING: Potentially large data request here
+                download_feather(root + filename, path, headers=self.dataset.header, overwrite=False)
 
         return sidecars
 
--- a/nomic/data_operations.py
+++ b/nomic/data_operations.py
@@ -440,7 +440,7 @@ class AtlasMapEmbeddings:
         Returns the path to downloaded embeddings.
         """
         self.projection.tile_destination.mkdir(parents=True, exist_ok=True)
-        root = f"{self.dataset.atlas_api_path}/v1/project/{self.dataset.id}/index/projection/{self.projection.id}/quadtree/"
+        root_url = f"{self.dataset.atlas_api_path}/v1/project/{self.dataset.id}/index/projection/{self.projection.id}/quadtree/"
 
         registered_sidecar_names = [sidecar[1] for sidecar in self.projection._registered_sidecars()]
         assert "embeddings" in registered_sidecar_names, "Embeddings not found in sidecars."
--- a/nomic/data_operations.py
+++ b/nomic/data_operations.py
@@ -1,4 +1,3 @@
-from atexit import register
 import base64
 import concurrent
 import concurrent.futures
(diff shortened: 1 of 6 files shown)


nomic/utils.py
```diff
@@ -241,10 +242,29 @@ def get_object_size_in_bytes(obj):
...
+    """
+    assert retries > 0, "Retries must be greater than 0"
+    download_attempt = 0
+    download_success = False
+    schema = None
+    while download_attempt < retries and not download_success:
+        download_attempt += 1
+        if not path.exists() or overwrite:
+            data = requests.get(url, headers=headers)
+            readable = BytesIO(data.content)
```
apage43, line 258: What is the largest file we expect to pull down here?

setup.py
```diff
@@ -43,7 +43,7 @@
         ],
         "dev": [
             "nomic[all]",
-            "black",
+            "black==24.3.0",
```
apage43, line 46: what's the reason for pinning this?

nomic/utils.py
rguo123: according to @bmschmidt, several hundreds of MB at max. Is this size reasonable?

setup.py
rguo123: there was a version bump and it changed some files that I didn't modify - may make it more difficult to review the actual changes for a more complex PR. Also thought it would be nice to match the version in atlas cloud but fine with keeping it unpinned

nomic/utils.py
apage43: mainly we have to buffer the whole file in memory (twice) working this way - a couple hundred MB is _probably_ okay

setup.py
apage43: ah, if its pinned in atlas_cloud we can pin it here, mostly I'm worried about forgetting we've pinned it

nomic/utils.py
apage43: I think using `pa.BufferReader` instead of `BytesIO` here can avoid the second copy (`read_table` has to use the file-like-object interface to read a BytesIO and therefore has to copy it, but a `BufferReader` can be handled zero-copy)

nomic/data_operations.py
```diff
@@ -426,53 +423,39 @@ def latent(self) -> np.ndarray:
...
-                feather.write_feather(tb, dest)
-                last = tilename
-                pbar.update(1)
+        downloaded_files_in_tile_order = []
+        logger.info("Downloading latent embeddings...")
+        all_quads = list(self.projection._tiles_in_order(coords_only=True))
+        for quad in tqdm(all_quads):
+            quad_str = os.path.join(*[str(q) for q in quad])
+            filename = quad_str + "." + "embeddings" + ".feather"
+            path = self.projection.tile_destination / Path(filename)
```
bmschmidt, line 454: Don't spend extra lines avoiding pathlib here:
```suggestion
        all_quads = list(self.projection._tiles_in_order())
        for base_tile in tqdm(all_quads):
            path = base_tile.with_suffix(".embeddings.feather")
```

nomic/data_operations.py
```diff
@@ -426,53 +423,39 @@ def latent(self) -> np.ndarray:
...
-        Downloads the latent embeddings one file at a time.
+        Downloads the feather tree for embeddings.
+        Returns the path to downloaded embeddings.
         """
-        logger.warning("Downloading latent embeddings of all datapoints.")
-        limit = 10_000
-        route = self.projection.dataset.atlas_api_path + "/v1/project/data/get/embedding/paged"
-        last = None
+        self.projection.tile_destination.mkdir(parents=True, exist_ok=True)
+        root = f"{self.dataset.atlas_api_path}/v1/project/{self.dataset.id}/index/projection/{self.projection.id}/quadtree/"
```
bmschmidt, line 443:
```suggestion
        root_url = f"{self.dataset.atlas_api_path}/v1/project/{self.dataset.id}/index/projection/{self.projection.id}/quadtree/"
```

nomic/data_operations.py
```diff
@@ -426,53 +423,39 @@ def latent(self) -> np.ndarray:
...
-                pbar.update(1)
+        downloaded_files_in_tile_order = []
+        logger.info("Downloading latent embeddings...")
+        all_quads = list(self.projection._tiles_in_order(coords_only=True))
+        for quad in tqdm(all_quads):
+            quad_str = os.path.join(*[str(q) for q in quad])
+            filename = quad_str + "." + "embeddings" + ".feather"
+            path = self.projection.tile_destination / Path(filename)
+            # WARNING: Potentially large data request here
+            download_feather(root + filename, path, headers=self.dataset.header, overwrite=False)
```
bmschmidt, line 456:
```suggestion
            download_feather(Path(root_url, *path.parts[-3:]), path, headers=self.dataset.header, overwrite=False)
```

nomic/utils.py
bmschmidt: No: the dance is to save user hard-drive space. 

Javascript doesn't support record batch compression. So our wire format is feather with no internal record-batch compression, wrapped in `Content-Encoding: gzip` so it gets quietly compressed/decompressed by the browser or, here, `requests` or whatever.

`pa.feather.write_feather` here will write it with `snappy` or something.

bmschmidt: We could probably kick the can down the road and write the bytes directly with the filename "z/x/y.embeddings.feather.gz" and do `feather.read_table` on `gzip.open` around the file, but that tends to be a lot slower and preclude mem-mapping them.

nomic/data_operations.py
rguo123: LOL yeah that's my bad. that is not a great way of adding a suffix hahahaha. Going to assign partial blame to copilot here

apage43 (approved):
good with this assuming its been tested on a decent sized dataset

PR #305, opened 2024-06-04
[RFR] Image embedding rgb and url support

rguo123:
Option to pass static URLs + rgb for jpg conversion

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,7 +8,7 @@ description = "The official Nomic python client."
 
 setup(
     name="nomic",
-    version="3.0.30",
+    version="3.0.31",
     url="https://github.com/nomic-ai/nomic",
     description=description,
     long_description=description,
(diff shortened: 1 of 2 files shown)


rguo123:
@zanussbaum think this is good to go!

ran
image(images=[image_url], model="nomic-embed-vision-v1.5") on a static url of a sushi image

and got this:
https://atlas.nomic.ai/data/nomic-multimodal-series/cc3m-100k-image-bytes/map/144e9b38-97b7-4c6f-9de3-8eeac2c5e731#epGf

setup.py
```diff
@@ -8,7 +8,7 @@
 
 setup(
     name="nomic",
-    version="3.0.29",
+    version="3.0.30",
```
zanussbaum, line 11: i think this needs to be 3.0.31, can you merge from main?

zanussbaum (approved):
LGTM, just make sure you merge from main. I think the version needs to be bumped one more

setup.py
rguo123: ah thanks for the catch, rebased and bumped

PR #307, opened 2024-06-06
[wip] Download refactor

rguo123:


rguo123:
PR is almost ready but blocked by quadtree manifest backfill

AndriyMulyar:
@ellipsis-dev review this

nomic/data_operations.py
```diff
@@ -93,35 +128,69 @@ def __init__(self, projection: "AtlasProjection"):  # type: ignore
...
+    def _load_topics(self):
+        """
+        Loads topics from the feather tree.
+        """
+        integer_topics = False
+        # pd.Series to match pd typing
+        label_df: Optional[Union[pd.DataFrame, pd.Series]] = None
+        if "int" in self._topic_columns[0][0]:
+            integer_topics = True
+            label_df = self.metadata[["topic_id", "depth", "topic_short_description"]]
```
bmschmidt, line 147: IMO it would be ok to explicitly, "temporarily" deprecate topic download on resources without the newer `_int` form.

nomic/data_operations.py
```diff
@@ -30,22 +26,57 @@ class AtlasMapDuplicates:
...
+            if field.startswith("_duplicate_class")
+        ]
+        cluster_columns = [
+            (field, sidecar) for field, sidecar in self.projection._registered_columns if field.startswith("_cluster")
+        ]
+
+        assert len(duplicate_columns) > 0, "Duplicate detection has not yet been run on this map."
+
+        self.duplicate_column = duplicate_columns[0]
+        self.cluster_column = cluster_columns[0]
```
bmschmidt, line 42: Would prefer these to be private (underscore) namespaced.

nomic/data_operations.py
```diff
@@ -260,8 +333,8 @@ def get_topic_density(self, time_field: str, start: datetime, end: datetime):
             A list of `{topic, count}` dictionaries, sorted from largest count to smallest count.
         """
         data = AtlasMapData(self.projection, fields=[time_field])
-        time_data = data._tb.select([self.id_field, time_field])
-        merged_tb = self._tb.join(time_data, self.id_field, join_type="inner").combine_chunks()
+        time_data = data.tb.select([self.id_field, time_field])
+        merged_tb = self.tb.join(time_data, self.id_field, join_type="inner").combine_chunks()
```
bmschmidt, line 337: Why this change?

nomic/utils.py
```diff
@@ -243,30 +243,48 @@ def get_object_size_in_bytes(obj):
 # Helpful function for downloading feather files
 # Best for small feather files
 def download_feather(
-    url: Union[str, Path], path: Path, headers: Optional[dict] = None, retries=3, overwrite=False
+    url: Union[str, Path], path: Path, headers: Optional[dict] = None, num_attempts=3, overwrite=False
```
bmschmidt, line 246: Should we just remove auto-retry on this? It seems like a strange feature to include in a library like this. If a download fails, it fails. They can retry now.

setup.py
```diff
@@ -8,7 +8,7 @@
 
 setup(
     name="nomic",
-    version="3.0.32",
+    version="3.1.0",
```
bmschmidt, line 11: Is the public API actually identical after this, or will things be different?

bmschmidt: Relatedly, is there a release_notes bump we can do here?

bmschmidt (approved):

nomic/data_operations.py
rguo123: We already had backwards compatibility built in when i first implemented it so doesnt feel like we have to remove

setup.py
rguo123: Hmm good point, i'm not sure so i'll just do 3.033 for now. I'm guessing a lot of this may change with lance so we can do a major bump then.

rguo123: Ideally, none of this code should really change how the user is using nomic client. Hopefully, it just makes the package more stable

nomic/data_operations.py
rguo123: done

nomic/data_operations.py
rguo123: We used to guarantee _tb was loaded when object was constructed because of the prefetch tiles logic. Now that it's removed, the download step only comes up if you call the tb property

nomic/utils.py
rguo123: hmm, let's keep it for now, but I can bump num_attempts down to 1.

PR #313, opened 2024-07-08
added connector folder and HF file

abhisomala:
HF file takes in any Huggingface identifier and then returns an AtlasDataset

### DIFF
--- a/connectors/huggingface_connecter.py
+++ b/connectors/huggingface_connecter.py
@@ -1,0 +1,46 @@
+from datasets import load_dataset
+import hashlib
+
+# Gets data from a Hugging Face dataset with automatic configuration
+def fetch_data_from_huggingface(dataset_identifier):
+    try:
+        # Try loading the dataset without specifying a configuration
+        dataset = load_dataset(dataset_identifier, trust_remote_code=True)
+    except ValueError as e:
+        # If there is an error it might be because of the config selection
+        if "Please pick one among the available configs" in str(e):
+            # Gets available config and selects first one
+            available_configs = str(e).split("['")[1].split("']")[0].split("', '")
+            dataset = load_dataset(dataset_identifier, available_configs[0], trust_remote_code=True)
+        else:
+            raise e
+
+    data = []
+    for split in dataset.keys():
+        for i, example in enumerate(dataset[split]):
+            # Creates a unique and shortened ID
+            unique_str = f"{dataset_identifier}_{split}_{i}"
+            short_id = hashlib.sha1(unique_str.encode()).hexdigest()[:25]
+            example['id'] = short_id
+            data.append(example)
+    return data
+
+# Main load function to be used as a connector
+def load(dataset_identifier):
+    data = fetch_data_from_huggingface(dataset_identifier.strip())
+
+    if data:
+        return data
+    else:
+        raise ValueError("No data was found for the provided dataset.")
+if __name__ == "__main__":
+    dataset_identifier = input("Enter Hugging Face dataset identifier: ").strip()
+
+    try:
+        data = load(dataset_identifier)
+        print(f"Dataset has been loaded successfully. Number of entries: {len(data)}")
+        # You can add more processing logic here if needed
+    except ValueError as e:
+        print(f"Error loading dataset: {e}")
+
+


connectors/huggingface_connecter.py
```diff
@@ -0,0 +1,57 @@
+from datasets import load_dataset
+import hashlib
+
+# Fetches data from a Hugging Face dataset
+def fetch_data_from_huggingface(dataset_identifier):
+    try:
+        # Attempts to load dataset without specifying config
+        dataset = load_dataset(dataset_identifier)
+    except ValueError as e:
```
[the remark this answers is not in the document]
apage43: See https://huggingface.co/docs/datasets/en/load_hub#splits

connectors/huggingface_connecter.py
RLesser, line 1: small nit but should be `huggingface_connector.py`

connectors/huggingface_connecter.py
```diff
@@ -0,0 +1,77 @@
...
+                entry[key] = ' '.join(map(str, value))
+            elif isinstance(value, np.ndarray):
+                entry[key] = ' '.join(map(str, value.flatten()))
+            elif hasattr(value, 'tolist'):
+                entry[key] = ' '.join(map(str, value.tolist()))
+            else:
+                entry[key] = str(value)
+
+
+    dataset.add_data(data=data)
```
RLesser, line 69: are we making an index here?

connectors/huggingface_connecter.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+    dataset = AtlasDataset(
+        map_name,
+        unique_id_field="id",
+    )
+
+
+    # Convert all booleans and lists to strings
+    for entry in data:
```
RLesser, line 55: two potential issues here:
1. This seems like it would be extremely slow (and possibly crash) for large huggingface datasets.
2. I'm a bit worried about assuming this is how people would want to handle these fields, but i guess they can edit it themselves if they want it done differently...

connectors/huggingface_connecter.py
```diff
@@ -0,0 +1,77 @@
...
+                dataset = load_dataset(dataset_identifier, available_configs[0], trust_remote_code=True)
+            except ValueError:
+                raise ValueError("Failed to get available configurations")
+        else:
+            raise e
+
+
+    # Processes dataset entries
+    data = []
+    for split in dataset.keys():
```
RLesser, line 28: same question here about large datasets - are we sure this will not break?

RLesser (commented):
some comments, maybe some things to change

apage43:
> Loads dataset using load_dataset library and assigns a unique ULID ID for each entry

> (andriy) These should be ULID not sh256 or uuids. @apage43 has a nice function for these

The primary reason to avoid pure-random IDs or hash based IDs is that they cause worst case performance when used as keys in ordered data structures (such b-tree indexes in a database), ULIDs improve on this by making the beginning of the ID a timestamp so that IDs created around the same time have some locality to each other, but, like UUIDs, they are still _kinda big_

Big (semi)random IDs like ULID are best used when you need uniqueness while also avoiding coordination, e.g. you have multiple machines inserting data into a collection and it would add a lot of complexity to make them cooperate to assign non-overlapping IDs - but in situations you where _can_ use purely sequential IDs it is usually better to, as smaller ids are cheaper to store *and* look up

When using `map_data` the nomic client already has functionality to create an sequential ID field (its still required to be a string so it base64s their binary representation), it may make sense to copy that behavior. See here: https://github.com/nomic-ai/nomic/blob/1f042befc53892271bd0a0877070d47b2d3cb631/nomic/atlas.py#L77

---

> Tested with datasets smaller than 10k for speed but can work with larger datasets

I believe this will not currently work when the dataset size exceeds available RAM on the machine running this - HF datasets understands slice syntax when specifying a split so you can test with portions of a very large dataset with `load_dataset("really-big-dataset", split="train[:100000]")` to only get the first 100k rows.

Making it work should be possible by working in chunks and using IterableDatasets https://huggingface.co/docs/datasets/v2.20.0/en/about_mapstyle_vs_iterable#downloading-and-streaming

here is a notebook where I'm uploading from an iterabledataset in chunks (note, though, that because I call load_dataset and then to_iterable_dataset this still downloads the _entire dataset_ - you can also pass `streaming=True` to load_dataset to get an IterableDataset that only downloads as much as you actually read, which may be desirable if you're only working with a subset of a large dataset): https://gist.github.com/apage43/9e80b0f4378ed466ec5d1c0a4042c398

RLesser:
Going off @apage43's comment, I feel strongly that we should be taking advantage of huggingface datasets use of Arrow to pass data to atlas, which also speaks fluent arrow. We should also be taking advantage of batching or chunking for arbitrarily large datasets. Using base python iterators means this will break for larger datasets.

connectors/huggingface_connecter.py
abhisomala: Not quite; that part mainly just converts anything that isn't a string, like booleans or lists, into strings because it will error out if I don't. I didn't really like how it is just a bunch of conditional statements but I couldn't find a _better_ way to resolve it.

connectors/huggingface_connecter.py
abhisomala: Could I solve this with batch processing and by passing `streaming=True` through load dataset? Aaron mentioned that so it seems like that could prevent that issue

connectors/huggingface_connector.py
```diff
@@ -0,0 +1,94 @@
...
+        return ' '.join(map(str, value.tolist()))
+    else:
+        return str(value)
+
+
+# Processes Arrow table and converts necessary fields to strings
+def process_table(table):
+    # Converts columns with complex types to strings
+    for col in table.schema.names:
+        column = table[col].to_pandas()
```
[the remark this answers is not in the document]
abhisomala: My code seems to throw an Attribute error without it. There are some cases that it works but with some like this one it doesn't. 
https://huggingface.co/datasets/Anthropic/hh-rlhf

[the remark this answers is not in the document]
apage43: can leave this as is for now then

apage43:
current version of this has no create_index calls so it'll only create an AtlasDataset with data in it but no *map* - is that intended?

connectors/huggingface_connector.py
```diff
@@ -0,0 +1,90 @@
+from datasets import load_dataset, get_dataset_split_names
+from nomic import AtlasDataset
+import pyarrow as pa
+import pyarrow.compute as pc
+
+
+# Gets data from HF dataset
+def get_hfdata(dataset_identifier, split="train", limit=100000):
+    try:
```
eelegiap, line 9: Add docstring to function definition

connectors/huggingface_connector.py
```diff
@@ -0,0 +1,90 @@
...
+        # Determine available splits
+        splits = get_dataset_split_names(dataset_identifier)
+        if split not in splits:
+            # Use the first available split if the specified split is not found
+            split = splits[0]
+
+
+        # Load the dataset
+        dataset = load_dataset(dataset_identifier, split=split, streaming=True)
```
eelegiap, line 18: Move `load_dataset` outside of try/except

connectors/huggingface_connector.py
```diff
@@ -0,0 +1,90 @@
...
+    except ValueError as e:
```
eelegiap, line 19: We don't need to handle this error, seems like handling config through the error message is too risky

eelegiap: So you can take out the try/except here

connectors/huggingface_connector.py
```diff
@@ -0,0 +1,90 @@
...
+            available_configs = str(e).split("['")[1].split("']")[0].split("', '")
+            dataset = load_dataset(dataset_identifier, available_configs[0], split=f"{split}[:{limit}]", trust_remote_code=True)
+        else:
+            raise e
+
+
+    # Processes dataset entries using Arrow
+    id_counter = 0
+    data = []
+    if dataset:
```
eelegiap, line 32: if no dataset exists, function should fail

connectors/huggingface_connector.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+# Function to convert complex types to strings using Arrow
+def process_table(table):
+    # Converts columns with complex types to strings
+    for col in table.schema.names:
+        column = table[col]
+        if pa.types.is_boolean(column.type):
+            table = table.set_column(table.schema.get_field_index(col), col, pc.cast(column, pa.string()))
+        elif pa.types.is_list(column.type):
```
eelegiap, line 52: I think if you flatten the column as a list then cast as a string, the column will be too long and not match up with the other columns of the table. We may need to refactor this slightly by making sure column length is the same and structs are handled on a row by row basis

examples/HF_example_usage.py
```diff
@@ -0,0 +1,26 @@
+import argparse
+from huggingface_connector import hf_atlasdataset
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description='Create an AtlasDataset from a Hugging Face dataset.')
+    parser.add_argument('--dataset_identifier', type=str, required=True, help='The Hugging Face dataset identifier')
```
eelegiap, line 7: you may want to explicitly specify in the arg that it's a hf_dataset_identifier because atlas has its own datasets

eelegiap (changes requested):

PR #355, opened 2024-11-12
fix: prevent autogenerated __index_level_0__ column uploading pandas dataframes

mcembalest:
Need to prevent the column name `__index_level_0__` from being auto-generated when a user uploads dataframe

In response to this problem from a user on discord:
- https://discord.com/channels/1076964370942267462/1076964911034400830/1305934851068792852

### DIFF
--- a/nomic/dataset.py
+++ b/nomic/dataset.py
@@ -15,6 +15,7 @@ import numpy as np
 import pyarrow as pa
 import requests
 from loguru import logger
+import pandas as pd
 from pandas import DataFrame
 from PIL import Image
 from pyarrow import compute as pc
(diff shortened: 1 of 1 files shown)


nomic/dataset.py
```diff
@@ -1373,6 +1373,10 @@ def add_data(
             blobs: A list of image paths, bytes, or PIL Images. Use if you want to create an AtlasDataset using image embeddings over your images. Note: Blobs are stored locally only.
             pbar: (Optional). A tqdm progress bar to update.
         """
+        if isinstance(data, DataFrame):
+            logger.info("DataFrame index was reset with pandas.DataFrame.reset_index() during upload to Atlas")
```
bmschmidt, line 1377: Can we check in pandas if there is actually an index?

bmschmidt: Or not log.

bmschmidt (approved):

PR #361, opened 2024-12-02
Normalize dataset slugs before sending request to server side

wilson-nomic:
Currently, our API normalizes the slug name, but the client is unaware of this normalization. This means that if we use a non-normal slug name that does exist normalized, the client will see that it doesn't exist, and try to create it, creating repeated duplicates.

Quick test:

# create.py

```python
from nomic.dataset import AtlasDataset

AtlasDataset("Foo Bar", unique_id_field="id")
```

# Before

```bash
(.venv) @wilson-nomic ➜ /workspaces/nomic (normalize-slug) $ python create.py
2024-12-02 20:39:15.563 | INFO     | nomic.dataset:_create_project:847 - Organization name: `wilsonlin`
2024-12-02 20:39:16.415 | INFO     | nomic.dataset:_create_project:875 - Creating dataset `foo-bar`
(.venv) @wilson-nomic ➜ /workspaces/nomic (normalize-slug) $ python create.py
2024-12-02 20:39:15.563 | INFO     | nomic.dataset:_create_project:847 - Organization name: `wilsonlin`
2024-12-02 20:39:16.415 | INFO     | nomic.dataset:_create_project:875 - Creating dataset `foo-bar-1`
(.venv) @wilson-nomic ➜ /workspaces/nomic (normalize-slug) $ python create.py
2024-12-02 20:39:15.563 | INFO     | nomic.dataset:_create_project:847 - Organization name: `wilsonlin`
2024-12-02 20:39:16.415 | INFO     | nomic.dataset:_create_project:875 - Creating dataset `foo-bar-2`
(.venv) @wilson-nomic ➜ /workspaces/nomic (normalize-slug) $ python create.py
2024-12-02 20:39:15.563 | INFO     | nomic.dataset:_create_project:847 - Organization name: `wilsonlin`
2024-12-02 20:39:16.415 | INFO     | nomic.dataset:_create_project:875 - Creating dataset `foo-bar-3`
```

# After

```bash
(.venv) @wilson-nomic ➜ /workspaces/nomic (normalize-slug) $ python create.py
2024-12-02 20:40:41.596 | INFO     | nomic.dataset:_create_project:847 - Organization name: `wilsonlin`
2024-12-02 20:40:42.420 | INFO     | nomic.dataset:_create_project:875 - Creating dataset `foo-foo`
(.venv) @wilson-nomic ➜ /workspaces/nomic (normalize-slug) $ python create.py
[...]

PR #367, opened 2024-12-11
fix: handle identifiers with slash

wilson-nomic:
Don't normalize organization part of dataset identifier if provided. Fixes regression caused by #361.

----

> [!IMPORTANT]
> Fixes regression in `AtlasDataset.__init__()` to correctly handle dataset identifiers with slashes by preserving the organization part.
> 
>   - **Behavior**:
>     - Fixes regression in `AtlasDataset.__init__()` by correctly handling dataset identifiers with slashes.
>     - Splits identifier on slash and normalizes only the dataset part, preserving the organization part.
>     - Reconstructs identifier with organization part if present.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=nomic-ai%2Fnomic&utm_source=github&utm_medium=referral)<sup> for ba7f6dba2f5cd3d8ab05a67c19efd1fcaa0c6853. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/nomic/dataset.py
+++ b/nomic/dataset.py
@@ -773,10 +773,13 @@ class AtlasDataset(AtlasClass):
         assert identifier is not None or dataset_id is not None, "You must pass a dataset identifier"
         # Normalize identifier.
         if identifier is not None:
-            identifier = unicodedata.normalize("NFD", identifier)  # normalize accents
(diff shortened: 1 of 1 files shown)


wilsonjr (approved):
LGTM

PR #372, opened 2025-01-18
fix: fix spelling

bmschmidt:
'alembert' is an autogenerated name, and it reads weirdly to not have it be `d'alembert`. So let's replace him with his partner in crime*.

*where 'crime' is against canon law.


mcembalest (approved):
lgtmbert

PR #373, opened 2025-01-22
fix: ValueError returned when calling map_data on blobs and embeddings

mcembalest:
> [!IMPORTANT]
> Adds a check in `map_data()` to raise `ValueError` if both `blobs` and `embeddings` are provided, ensuring correct usage of parameters.
> 
>   - **Behavior**:
>     - In `map_data()` in `atlas.py`, added a check to raise `ValueError` if both `blobs` and `embeddings` are provided, clarifying that only one can be used at a time.
>   - **Error Handling**:
>     - Provides a specific error message guiding users on how to correctly use `blobs` and `embeddings` parameters.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=nomic-ai%2Fnomic&utm_source=github&utm_medium=referral)<sup> for befec4f597f37d359f9814db0e8f44b0c0452253. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/nomic/atlas.py
+++ b/nomic/atlas.py
@@ -64,7 +64,9 @@ def map_data(
 
     if blobs is not None:
         if embeddings is not None:
-            raise ValueError("You cannot pass both `blobs` and `embeddings` to map_data(). To create a map of images, include `blobs` and not `embeddings`. To create a map of embeddings with images as metadata, include your images as a field in your `data`.")
+            raise ValueError(
+                "You cannot pass both `blobs` and `embeddings` to map_data(). To create a map of images, include `blobs` and not `embeddings`. To create a map of embeddings with images as metadata, include your images as a field in your `data`."
+            )
         # change this when we support other modalities
         modality = "image"
         indexed_field = "_blob_hash"
--- a/nomic/atlas.py
+++ b/nomic/atlas.py
@@ -65,7 +65,7 @@ def map_data(
     if blobs is not None:
         if embeddings is not None:
             raise ValueError(
-                "You cannot pass both `blobs` and `embeddings` to map_data(). To create a map of images, include `blobs` and not `embeddings`. To create a map of embeddings with images as metadata, include your images as a field in your `data`."
+                "You cannot pass both `blobs` and `embeddings` to map_data(). To create a map of images, include `blobs` and not `embeddings`. To create a map of embeddings with images as metadata, include your images as a field in your `data` parameter."
             )
         # change this when we support other modalities
         modality = "image"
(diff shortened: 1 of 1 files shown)


bmschmidt (approved):

brendanashworth:
what's the correct way to provide our own embeddings when using image data then? It errors when I try to use a PIL Image in a `data` field. downgrading before this change still works great for my use case so I'm confused

mcembalest:
Hi @brendanashworth, what you're describing (uploading PIL images with custom embeddings) is not currently supported but is on our roadmap.

Currently the way to get images with custom embeddings in an Atlas map is to pass your embeddings to the `embeddings` parameter of `atlas.map_data` and your images **as URLS** in the `data` parameter, e.g.:

Screenshot 2025-02-11 at 12 21 49 PM

```python
atlas.map_data(
    identifier='custom-image-embeddings',
    embeddings=your_image_embeddings_array, # (N,d) array of embeddings
    data=df # pandas dataframe, pyarrow table, or list of dicts including a field for the image URL
)
```

This will render each image hosted at the given URLs in the sidebar when you are browsing your map in Atlas. 

Screenshot 2025-02-11 at 12 16 46 PM

brendanashworth:
Thank you @mcembalest , does this mean that base64-encoded images uploaded directly are no longer supported if I'm using my own embeddings? And I need to host the images myself? Giving the image blobs was very convenient for me.

brendanashworth:
@mcembalest it seems as though passing base64-encoded URLs doesn't work:

Screenshot 2025-02-11 at 11 25 39 AM

mcembalest:
> @mcembalest it seems as though passing base64-encoded URLs doesn't work:
> 
> Screenshot 2025-02-11 at 11 25 39 AM

@brendanashworth OK you should be able to upload images as base64 text this way!

```python
from PIL import Image
import base64
from io import BytesIO
def pil_to_img_tag(image: Image.Image) -> str:
    """Converts a PIL image to an img div with the image in base64"""
    buffer = BytesIO()
    image.save(buffer, format="PNG")
    base64_data = base64.b64encode(buffer.getvalue()).decode("utf-8")
    return f"<img src='data:image/png;base64,{base64_data}'>"

# add an 'image' field to df
# using pil_to_img_tag helper function
df['image'] = [
    pil_to_img_tag(img) for img in images
]

# create map with custom embeddings
atlas.map_data(
    identifier='custom-image-embeddings',
    embeddings=your_image_embeddings_array, # (N,d) array of embeddings
    data=df # pandas dataframe including a field for the images encoded as base64 in a div
)
```

brendanashworth:
Thanks @mcembalest , very helpful, that works!

